PR #55, opened 2014-09-23
make punycode decoding optional

strcrzy:
there exist domains that start with the `xn--` string or contain it at the beginning of a subdomain that are not in fact punycode-encoded domains. these domains choke tldextract.

this revision allows a tldextract user to direct tldextract to treat punycode-encoded domains as plain old strings.


john-kurkowski:
> these domains choke tldextract.

Does #67 address your original problem at all?

strcrzy:
yes, thank you :)

PR #61, opened 2015-02-11
Add possibility to pass extra_suffixes directly

shirk3y:


john-kurkowski:
Looks good!

I never thought anybody would reconcile the runtime suffixes with the cached ones. The existing caching code is a bit messy. You navigated it well.

Issue #19, opened 2012-10-16
Incorrect tld for appspot.com (support excluding private domains on lookup)


davefd:
There are multiple uses of the <a href="https://wiki.mozilla.org/Public_Suffix_List/Use_Cases">Mozilla  public suffix list</a> which allow sites such as "appspot.com" to appear on the list as a tld instead of being split into domain="appspot" and tld="com".

This is perfectly reasonable behavior for some use cases, but for others it would be helpful to have the "private" domains be excluded.  <a href="https://bugzilla.mozilla.org/show_bug.cgi?id=712640">Mozilla has split the list</a> into "ICANN Domains" and "Private Domains", and it would be useful to optionally be able to exclude the private domains so that sites like "appspot.com" would have their tld reflected as "com".

PR #65, opened 2015-03-05
tldextract --update can be configured to download private domains

### LINKED ISSUE #64: tldextract --update does not download private domains
If I run tldextract --update to update domains file then, because of this line https://github.com/john-kurkowski/tldextract/blob/1.5.1/tldextract/tldextract.py#L264 the private domains are not downloaded.
Therefore, if later I create TLDExtract(include_psl_private_domains=True) class, the private domains are not mapped.
I recommend always downloading the full list and replace line 264 as 
TLD_EXTRACTOR = TLDExtract(include_psl_private_domains=True)

kristjanr:
fixes #64

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -281,7 +281,7 @@ class TLDExtract(object):
             return frozenset(suffixes)
         return suffixes
 
-TLD_EXTRACTOR = TLDExtract()
+TLD_EXTRACTOR = TLDExtract(include_psl_private_domains=True)
 
 
 @wraps(TLD_EXTRACTOR.__call__)
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -281,7 +281,7 @@ class TLDExtract(object):
             return frozenset(suffixes)
         return suffixes
 
-TLD_EXTRACTOR = TLDExtract(include_psl_private_domains=True)
+TLD_EXTRACTOR = TLDExtract()
 
 
 @wraps(TLD_EXTRACTOR.__call__)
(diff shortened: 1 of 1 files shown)


tldextract/tldextract.py
```diff
@@ -281,7 +281,7 @@ def _add_extra_suffixes(self, suffixes):
             return frozenset(suffixes)
         return suffixes
 
-TLD_EXTRACTOR = TLDExtract()
+TLD_EXTRACTOR = TLDExtract(include_psl_private_domains=True)
```
john-kurkowski: I like where this is going. I agree we must persist all public/private domains, but only include private results if the user opts into them. However this edit will include private domains for anybody who uses the export `extract`, just a few lines down. That would undo #19.

kristjanr:
another try to fix #64

tldextract/tldextract.py
```diff
@@ -374,12 +374,13 @@ def main():
     parser.add_argument('-c', '--cache_file', help='use an alternate TLD definition file')
 
     args = parser.parse_args()
+    tld_extract = TLDExtract(include_psl_private_domains=True)
```
john-kurkowski: Anybody who runs `tldextract --update` and later uses the same cache file will be forced to have public/private domains in their results, even if they only want public domains.

Ideally, we store metadata in the persisted file so results can be filtered at runtime.

For the 1.x series though, I think I'd be happy with `include_psl_private_domains` being configurable here.

john-kurkowski: For the ideal fix, I opened #66.

john-kurkowski: > For the 1.x series though, I think I'd be happy with include_psl_private_domains being configurable here.

Bump. Sorry if it wasn't clear, this is the only blocker on this PR. Should be an easy update.

(I don't think anybody's gonna tackle #66 anytime soon. Don't hold your breath for it.)

kristjanr:
Huh?

john-kurkowski:
If you make `include_psl_private_domains` configurable, defaulting to false, I can accept this. Otherwise, the PR will inflict private domains on everybody.

john-kurkowski:
Thank you!

PR #67, opened 2015-03-17
Catching exceptions when calling .encode and .decode on punycode that isn't punycode

xdanx:
On line https://github.com/john-kurkowski/tldextract/blob/master/tldextract/tldextract.py#L191 `.encode` or `.decode` should catch at least `UnicodeError`. 

Example bad domain: `xn--tub-test.google.com` :

```
UnicodeError: ('IDNA does not round-trip', 'xn--tub-', 'tub')
```
I don't know exactly why this happens, I think it's something to do with the .decode function, but it's irrelevant to this project.

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -192,9 +192,7 @@ class TLDExtract(object):
             try:
                 netloc = codecs.decode(netloc.encode('ascii'), 'idna')
             except UnicodeError:
-                LOG.warn('UnicodeError when decoding netloc ' + netloc)
-                # We return the answer as it is
-                # We consider this not to be punycode!
+                # We return the answer as it is, not considering the input punycode
                 is_punycode = False
 
         registered_domain, tld = self._get_tld_extractor().extract(netloc)
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -192,7 +192,7 @@ class TLDExtract(object):
             try:
                 netloc = codecs.decode(netloc.encode('ascii'), 'idna')
             except UnicodeError:
-                LOG.error('UnicodeError when decoding netloc ' + netloc)
+                LOG.warn('UnicodeError when decoding netloc ' + netloc)
                 # We return the answer as it is
                 # We consider this not to be punycode!
                 is_punycode = False
@@ -200,19 +200,8 @@ class TLDExtract(object):
         registered_domain, tld = self._get_tld_extractor().extract(netloc)
 
         if is_punycode:
-            """
-            If we are here, means that:
-            1) netloc is punycode
-            2) netloc decoded successfully to unicode
-            As such, we'll try to encode the netloc to ascii
-            """
-            try:
-                registered_domain = codecs.encode(registered_domain, 'idna')
-                tld = codecs.encode(tld, 'idna')
-            except UnicodeError:
-                LOG.error('UnicodeError when encoding netloc ' + netloc)
-                # If we get an error here, we give up
-                return ExtractResult('', '', '')
+            registered_domain = codecs.encode(registered_domain, 'idna')
+            tld = codecs.encode(tld, 'idna')
 
         if not tld and netloc and netloc[0].isdigit():
             try:
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -187,14 +187,32 @@ class TLDExtract(object):
             .rstrip(".")
 
         is_punycode = netloc.startswith('xn--') or '.xn--' in netloc
+        # Some IDN urls might generate an error
         if is_punycode:
-            netloc = codecs.decode(netloc.encode('ascii'), 'idna')
+            try:
+                netloc = codecs.decode(netloc.encode('ascii'), 'idna')
+            except UnicodeError:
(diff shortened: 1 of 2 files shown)


tldextract/tldextract.py
```diff
@@ -187,14 +187,32 @@ def __call__(self, url):
...
+                LOG.error('UnicodeError when decoding netloc ' + netloc)
+                # We return the answer as it is
+                # We consider this not to be punycode!
+                is_punycode = False
 
         registered_domain, tld = self._get_tld_extractor().extract(netloc)
 
         if is_punycode:
-            registered_domain = codecs.encode(registered_domain, 'idna')
-            tld = codecs.encode(tld, 'idna')
```
john-kurkowski: Why does this hunk have to change if it decoded fine on 191?

john-kurkowski:
I think the first hunk is on the right track! This library tends to be forgiving and silent about input. It doesn't expect validated URLs, punycode or otherwise.

john-kurkowski:
Can you add an accompanying test?

xdanx:
First of all, I did a mistake with my example domain, here are 2 problematic domains: 
```
xn--tub-1m9d15sfkkhsifsbqygyujjrw602gk4li5qqk98aca0w.google.com <= Does not round trip
xn--tub-1m9d15sfkkhsifsbqygyujjrw60.google.com <= Incomplete punycode string
```
2) My reasoning was that if .decode generates errors, .encode might do the same and I wanted to guard it with a try/catch, but I agree with you: if domain is indeed punycode and decodeds successfully, I *guess* it will also encode successfully. 

3) I don't understand what you mean by 
> I think the first hunk is on the right track! This library tends to be forgiving and silent about input. It doesn't expect validated URLs, punycode or otherwise.

As in should we not catch UnicodeErrors? Or that we should expect every input to be 'valid' ? At least in this case we should mention somewhere that some punycode domains might generate an exception. 

4) Added tests for the 2 domains I mentioned above. I might be able to catch other domains if you prefer to add more tests.

john-kurkowski:
> As in should we not catch UnicodeErrors? Or that we should expect every input to be 'valid' ? At least in this case we should mention somewhere that some punycode domains might generate an exception.

Oops, all I meant was, I agree with this PR's approach.

tldextract/tldextract.py
```diff
@@ -187,8 +187,15 @@ def __call__(self, url):
             .rstrip(".")
 
         is_punycode = netloc.startswith('xn--') or '.xn--' in netloc
+        # Some IDN urls might generate an error
         if is_punycode:
-            netloc = codecs.decode(netloc.encode('ascii'), 'idna')
+            try:
+                netloc = codecs.decode(netloc.encode('ascii'), 'idna')
+            except UnicodeError:
+                LOG.warn('UnicodeError when decoding netloc ' + netloc)
```
john-kurkowski: One final nitpick: is this log necessary? This hunk is `tldextract`'s attempt to interpret the input as punycode. If `tldextract` guessed wrong, that's on the library, not the user. The user doesn't need to know. Similar to the `try`/`except` in lines 207-214.

xdanx: Uhm, no it's not. I'm used to log every "unintended message" but I agree with you. So I'll remove it.

john-kurkowski: Cool!

I understand where you're coming from. Outside this library, I don't like failing silently either. I like preventing errors before they can even occur.

But it's this library's stance that, there's plenty of URL validators out there. Rather than inflict their overhead on all users, if it's important, they can opt into validation themselves. And it seems to be Python's idiom to avoid typechecks, be adults, RTFM (often, RTFS), and pass expected args. Keeps this library light and easy to get started with. I will document this design approach better.

I do wonder if this lenient approach has desensitized users to the nuances of URLs in the wild...

john-kurkowski:
Thank you very much!

tldextract/tldextract.py
john-kurkowski: By the way, that rant above had nothing to do with the PR really. I was just thinking out loud. Documented that thinking in the FAQ in 12f43eb3b3c94f3dcf071b4d826fdc609f4f00cf.

Again, thanks for the fix!

xdanx: Yes, I agree with you as well. 

I also support the RTFM / RTFS and force users pass the expected args. Unfortunately, I also don't like when my python scripts crash with no apparent reason and when I stumbled upon those weird domains, it made tldextract crash. 

Problem is a bit more "complex" because a domain that starts with xn-<blabla> can be either a punycode or just .. .a normal domain! Punycode domains are"valid" as long as users want them to be "valid". From this lib's point of view, we don't care if a punycode subdomain is valid or not, we just need to split them. But here comes the problem... in order to try to match a punycode domain from PSL we need to... convert it if possible. At least we need to try. And if we fail.. then bad luck, it wasn't punycode :) But our job is still to split that URL.Hence the need to try/catch conversion. 

Just wanted to clarify what is my reasoning in case there will be future problems :)

xdanx:
Glad I could help! Thank you for your quick merge. If I run into other bugs in the future I'll let you know!

strcrzy:
thank you!

PR #69, opened 2015-05-23
On Python 3+, the ExtractResult namedtuples __dict__ attribute was not populated

otakucode:
Fixing that and adding unit test to exercise issue.


john-kurkowski:
Good catch. Thanks!

PR #73, opened 2015-06-11
Mixed puny

maxmzkr:
I'm not sure if you want this in the code base. It came to mind so I thought I'd write the code and then open a dialog with you.

Take the following url for example, xn--zckzap6140b352b.blog.so-net.大分.jp. The code would currently convert all of it to unicode, then convert all of it back to puny causing the unicode text to not be returned the way it was presented. Keeping an index on the labels that were originally puny code takes care of this.

This pull request includes the other pull request I submitted as there was no way to do this one without the other.

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -189,14 +189,19 @@ class TLDExtract(object):
             .partition(":")[0] \
             .rstrip(".")
 
-        is_punycode = netloc.startswith('xn--') or '.xn--' in netloc
-        # Some IDN urls might generate an error
-        if is_punycode:
-            try:
-                netloc = codecs.decode(netloc.encode('ascii'), 'idna')
-            except UnicodeError:
-                # We return the answer as it is, not considering the input punycode
-                is_punycode = False
+        unicode_labels = []
+        is_punycode = False
+        for label in netloc.split("."):
+            tmp_is_punycode = netloc.startswith('xn--')
+            if tmp_is_punycode:
+                try:
+                    label = codecs.decode(label.encode('ascii'), 'idna')
+                    is_punycode = True
+                except UnicodeError:
+                    pass
+
+            unicode_labels.append(label)
+        netloc = ".".join(unicode_labels)
 
         registered_domain, tld = self._get_tld_extractor().extract(netloc)
 
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -189,31 +189,24 @@ class TLDExtract(object):
             .partition(":")[0] \
             .rstrip(".")
 
-        unicode_labels = []
-        puny_index = []
-        for label in netloc.split("."):
-            is_punycode = label.startswith('xn--')
-            if is_punycode:
+        labels = netloc.split(".")
+        translations = []
+        for label in labels:
+            if label.startswith("xn--"):
                 try:
-                    label = codecs.decode(label.encode('ascii'), 'idna')
+                    translation = codecs.decode(label.encode('ascii'), 'idna')
                 except UnicodeError:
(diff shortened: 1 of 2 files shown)


tldextract/tldextract.py
```diff
@@ -336,6 +347,24 @@ def _decode_utf8(s):
...
+    unicode label to puny if it's corresponding puny index is True
+
+    >>> merge_to_puny(
+    ...     [u"\u7e54\u572d\u30c6\u30cb\u30b9", u"blog", u"so-net",
+    ...      u"\u7e54\u572d\u30c6\u30cb", "jp"],
+    ...     [True, False, False, False, False])
+    u'xn--zckzap6140b352b.blog.so-net.\u7e54\u572d\u30c6\u30cb.jp'
+    """
+    puny_labels = []
+    for unicode_label, is_puny in zip(unicode_labels, puny_index):
```
john-kurkowski: Rather than juggling 2 lists, maybe a single dict mapping string to is_puny?

tldextract/tldextract.py
```diff
@@ -189,20 +189,31 @@ def __call__(self, url):
...
+            puny_index.append(is_punycode)
+        netloc = ".".join(unicode_labels)
 
         registered_domain, tld = self._get_tld_extractor().extract(netloc)
 
-        if is_punycode:
-            registered_domain = codecs.encode(registered_domain, 'idna').decode('utf-8')
-            tld = codecs.encode(tld, 'idna').decode('utf-8')
+        num_domain_labels = registered_domain.count(".")
+        if any(puny_index[:num_domain_labels]):
```
john-kurkowski: What is this checking? Would it be clearer as a self-documenting variable?

tldextract/tldextract.py
```diff
@@ -189,20 +189,31 @@ def __call__(self, url):
...
+            domain_labels = registered_domain.split(".")
+            registered_domain = merge_to_puny(domain_labels,
+                                              puny_index[:num_domain_labels])
+
+        if any(puny_index[num_domain_labels:]):
```
john-kurkowski: What is this checking? Would it be clearer as a self-documenting variable?

john-kurkowski:
Sorry it took me so long to look at this. I think this is great! Let's favor this over #72.

My only concern is it changes the performance characteristics with string splitting and joining, redundant with the rest of tldextract. But hey, I'm probably prematurely optimizing. We can give it a try as is.

I just had a couple inline questions. Other than that, good to merge.

maxmzkr:
Your concerns are valid. I profiled it over a sample of 10000 urls and found it to be about 10% slower. I'll take another pass and address you comments and see if I can do anything to help with performance.

maxmzkr:
I redid the algorithm and got much better results. I took 5 sample on each branch, running it over 1,000,000 urls.

Master
real: 14.2632
user: 14.0768
sys: 0.1776

Mixed Puny
real: 14.6516
user: 14.452
sys: 0.1896

Percent Increase
real: 2.7%
user: 2.7%
sys: 6.8%

john-kurkowski:
Wow, quick turnaround! And fantastic speed. That helps me sleep a lot better. ;)

john-kurkowski:
Curious, what did you do to profile?

maxmzkr:
I used a large sample set that the company I work for has. The sample set
is a list of DNS request that we have seen in the wild on large subscriber
networks. I took the top one million DNS names and ran an extract on them.
As for timing, I just used UNIX's time.

john-kurkowski:
Nice. I'll have to scrounge an open source version of that set for future profiles.

Thanks for your contribution!

PR #74, opened 2015-09-24
Support IPv6 addresses

mrmrwat:
IPv6 addresses within URLs must be wrapped in `[]` so we can partition on `]` once the netloc has been isolated.
If the IPv6 address is not part of a URL (or is not wrapped in brackets) it cannot have a port associated with it, so we can check for a single `:` and partition on that. If it has more than one colon it, it is likely to be an IPv6 address (unless it is just a malformed hostname string). I have added extra checking for IPv6 addresses in the section where IPv4 addresses are tested, however this is not strictly necessary since they would be placed in the domain location of an `ExtractResult` anyway.
I have also reordered the way partitioning of a URL is handled since it makes more sense to remove elements from right to left (i.e. fragment first, then query and path) until we are left with the netloc.

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -182,12 +182,15 @@ class TLDExtract(object):
         ExtractResult(subdomain='forums', domain='bbc', suffix='co.uk')
         """
         netloc = SCHEME_RE.sub("", url) \
-            .partition("/")[0] \
-            .partition("?")[0] \
             .partition("#")[0] \
+            .partition("?")[0] \
+            .partition("/")[0] \
             .split("@")[-1] \
-            .partition(":")[0] \
-            .rstrip(".")
+
+        if ']' in netloc:
+            netloc = netloc.partition("]")[0].strip("[")
+        elif netloc.count(':') == 1:
+            netloc = netloc.partition(":")[0]
 
         labels = netloc.split(".")
         translations = []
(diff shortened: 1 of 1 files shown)


tldextract/tldextract.py
```diff
@@ -208,15 +211,20 @@ def __call__(self, url):
...
-                is_ip = socket.inet_aton(netloc)
+                is_ip = socket.inet_pton(socket.AF_INET, netloc)
                 return ExtractResult('', netloc, '')
             except (AttributeError, UnicodeError):
                 if IP_RE.match(netloc):
                     return ExtractResult('', netloc, '')
             except socket.error:
                 pass
+            try:
+                is_ip = socket.inet_pton(socket.AF_INET6, netloc)
```
john-kurkowski: In your description, you're saying this `try` is unnecessary? Line 230 will do the right thing?

john-kurkowski:
Thanks for this! Makes sense. It breaks 2 tests though. And can you add new tests for your use case?

mrmrwat:
Hi John. No problem, although I'm a bit swamped at work, not sure when I'll be able to make a new commit. As for your comment on the code, I am using a version at work that doesn't bother checking the IPv6 address once it has been identified. Since it doesn't have a valid TLD and doesn't contain a subdomain it simply gets put into the domain field anyway, so I might remove that try/except.

john-kurkowski:
I would still be interested in IPv6 support! But closing this, due to no activity in several months. A lot has happened since.

mrmrwat:
Hi John,

Yeah sorry about that. I've been super busy with changing jobs and have
even had a baby since writing that patch! :)
I might have time in the future, but it's ok to stay closed for now.

Mike

Issue #75, opened 2015-11-04
Domain parsing fails with trailing spaces


evanv:
The text values passed to extract should strip the text for trailing spaces, eg, 

```
>>> tldextract.extract('ads.adiquity.com ')
ExtractResult(subdomain='ads.adiquity', domain='com ', suffix='')
>>> tldextract.extract('ads.adiquity.com   '.strip())
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
```

hangtwenty:
I think this is on you, not tldextract. 2 cents

evanv:
Perhaps. Lines are a bit blurry here for me. On the one hand sure it's the calling code's responsibility. On the other I doubt if my team is the first one that's stumbled on this, or that we'll be the last.  

In any event, that's certainly how we will respond to it. But either way I think it deserved to be mentioned here as an issue (perhaps as you say one that shouldn't be fixed)

hangtwenty:
I get you, I'm glad you opened the ticket. So. Assessing whether it should be changed.

I don't think other tools would do this cleaning for you either. TLDExtract expects a valid domain name to be given. A domain name with a space isn't valid. Note, for example, that you couldn't make a request to this URL.

```
>>> requests.get('http://ads.adiquity.com ')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python2.7/site-packages/requests/api.py", line 69, in get
    return request('get', url, params=params, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 468, in request
    resp = self.send(prep, **send_kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/sessions.py", line 576, in send
    r = adapter.send(request, **kwargs)
  File "/usr/local/lib/python2.7/site-packages/requests/adapters.py", line 423, in send
    raise ConnectionError(e, request=request)
requests.exceptions.ConnectionError: HTTPConnectionPool(host='ads.adiquity.com%20', port=80): Max retries exceeded with url: / (Caused by NewConnectionError('<requests.packages.urllib3.connection.HTTPConnection object at 0x10ad50890>: Failed to establish a new connection: [Errno 8] nodename nor servname provided, or not known',))
```

or aside from Python, let's try `curl`

```
[...]

PR #76, opened 2015-11-08
strip whitespace from urls

evanv:
closes https://github.com/john-kurkowski/tldextract/issues/75

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -187,6 +187,7 @@ class TLDExtract(object):
             .partition("#")[0] \
             .split("@")[-1] \
             .partition(":")[0] \
+            .strip() \
             .rstrip(".") \
             .strip()
 
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -294,7 +294,7 @@ TLD_EXTRACTOR = TLDExtract()
 
 @wraps(TLD_EXTRACTOR.__call__)
 def extract(url):
-    return TLD_EXTRACTOR(url)
+    return TLD_EXTRACTOR(url.strip())
 
 
 @wraps(TLD_EXTRACTOR.update)
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -188,7 +188,7 @@ class TLDExtract(object):
             .split("@")[-1] \
             .partition(":")[0] \
             .rstrip(".") \
-            .rstrip(" ")
+            .strip()
 
         labels = netloc.split(".")
         translations = []
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -188,8 +188,7 @@ class TLDExtract(object):
             .split("@")[-1] \
             .partition(":")[0] \
             .strip() \
-            .rstrip(".") \
-            .strip()
+            .rstrip(".")
 
         labels = netloc.split(".")
         translations = []
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -187,7 +187,8 @@ class TLDExtract(object):
             .partition("#")[0] \
             .split("@")[-1] \
             .partition(":")[0] \
-            .rstrip(".")
+            .rstrip(".") \
+            .rstrip(" ")
 
         labels = netloc.split(".")
         translations = []
@@ -294,7 +295,7 @@ TLD_EXTRACTOR = TLDExtract()
 
 @wraps(TLD_EXTRACTOR.__call__)
 def extract(url):
-    return TLD_EXTRACTOR(url.strip())
+    return TLD_EXTRACTOR(url)
 
 
 @wraps(TLD_EXTRACTOR.update)


tldextract/tldextract.py
```diff
@@ -294,7 +294,7 @@ def _add_extra_suffixes(self, suffixes):
 
 @wraps(TLD_EXTRACTOR.__call__)
 def extract(url):
-    return TLD_EXTRACTOR(url)
+    return TLD_EXTRACTOR(url.strip())
```
john-kurkowski: What about URLs passed to a `TldExtract` instance directly?

evanv:
@john-kurkowski you're absolutely right. My apologies. I believe this will do the job. When you have a second could you take a look? 

```
>>> import tldextract
>>> extract = tldextract.TLDExtract()
>>> extract('ads.adiquity.com')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
>>> extract('ads.adiquity.com ')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
>>> extract('ads.adiquity.com              ')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com’)
```

hangtwenty:
@evanv any reason why you decided to use `.rstrip()` instead of `.strip()`? Also, shouldn't just be `" "` as argument. I think it should be exactly `.strip()`. The title of #75 calls out "trailing spaces" but I don't see why we shouldn't strip whitespace on both sides.

evanv:
>any reason why you decided to use .rstrip() instead of .strip()

@hangtwenty nope!

I'm a SQL/Java guy, not really a Python one is basically the reason. My team uses it and I'm doing what I can to pick it up. And so I definitely appreciate feedback like you just shared. But the reason was ignorance/inexperience with Python on my behalf. Thanks for correcting me! 

```
>>> import tldextract
>>> extract = tldextract.TLDExtract()
>>> extract('ads.adiquity.com')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
>>> extract('ads.adiquity.com ')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
>>> extract(' ads.adiquity.com ')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
>>> extract('      ads.adiquity.com        ')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
>>> extract('      ads.adiquity.com')
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
```

evanv:
Also to the point of the issue at hand....  I 100% agree that it should trim whitespace on both sides. And when you mentioned it, I agreed immediately it should be strip not rstrip. Just wasn't thinking very carefully when I updated it! My mistake entirely!

hangtwenty:
No problem at all. Thanks for changing it :+1: 

Going to wait until tomorrow to see if we hear any other discussion on #75. Otherwise I think we are ready to merge :)

tldextract/tldextract.py
```diff
@@ -187,7 +187,8 @@ def __call__(self, url):
             .partition("#")[0] \
             .split("@")[-1] \
             .partition(":")[0] \
-            .rstrip(".")
+            .rstrip(".") \
+            .strip()
```
john-kurkowski: Should this go before the `rstrip` call? The trailing dot the `rstrip` is meant to take care of would not be stripped if there were whitespace in its way.

john-kurkowski:
Besides my comment, :+1:

mauricioabreu:
:+1:

evanv:
So this is actually interesting @john-kurkowski 

I wrote a little test to see. This was the test: https://gist.github.com/evanv/e27d553a1c536f62e6a3

The first time I ran it with the following rules:
```
netloc = SCHEME_RE.sub("", url) \
    .partition("/")[0] \
    .partition("?")[0] \
    .partition("#")[0] \
    .split("@")[-1] \
    .partition(":")[0] \
    .rstrip(".") \
    .strip()
```

The outcome was

```
$ python test.py
'ads.adiquity.com'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com.'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com.'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com. '
ExtractResult(subdomain='ads.adiquity.com', domain='', suffix='')
'ads.adiquity.com. '
ExtractResult(subdomain='ads.adiquity.com', domain='', suffix='')
'ads.adiquity.com .'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com .'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com . '
ExtractResult(subdomain='ads.adiquity.com ', domain='', suffix='')
'ads.adiquity.com '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
Number of correctly extracted subdomains: 7
Out of: 10
```

The second time I moved the strip to the inside of the rstrip, like you suggested. 
```
netloc = SCHEME_RE.sub("", url) \
    .partition("/")[0] \
    .partition("?")[0] \
    .partition("#")[0] \
    .split("@")[-1] \
    .partition(":")[0] \
    .strip() \
    .rstrip(".") 
```
Now I get
```
$ python test.py
'ads.adiquity.com'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com.'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com.'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com. '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com. '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com .'
ExtractResult(subdomain='ads.adiquity', domain='com ', suffix='')
' ads.adiquity.com .'
ExtractResult(subdomain='ads.adiquity', domain='com ', suffix='')
' ads.adiquity.com . '
ExtractResult(subdomain='ads.adiquity', domain='com ', suffix='')
'ads.adiquity.com '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
Number of correctly extracted subdomains: 7
Out of: 10
```

I actually want to test this a bit more but interestingly enough applying strip before and after

```
netloc = SCHEME_RE.sub("", url) \
    .partition("/")[0] \
    .partition("?")[0] \
    .partition("#")[0] \
    .split("@")[-1] \
    .partition(":")[0] \
    .strip() \
    .rstrip(".") \
    .strip()
```

appears to do a lot better
```
$ python test.py
'ads.adiquity.com'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com.'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com.'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com. '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com. '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com .'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com .'
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com . '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
'ads.adiquity.com '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
' ads.adiquity.com '
ExtractResult(subdomain='ads', domain='adiquity', suffix='com')
Number of correctly extracted subdomains: 10
Out of: 10    
```

I'll take another look after work today and try to assemble a better set of test urls.... also want to think through it a bit more carefully after I've had some coffee. Either way... I don't think moving strip inside of rstrip is quite right either... although calling strip before and after rstrip might be

hangtwenty:
@john-kurkowski I'm confused by this.

Using one of @evanv 's helpful examples, of a case where doing two `.strip()` calls works: `' ads.adiquity.com . '` Not sure this is a use-case we should support. As you know I'm sure ...

> It's a little-known fact, but fully-qualified (unambiguous) DNS domain names have a dot at the end. 

(Pulling a source out of a hat... http://www.dns-sd.org/trailingdotsindomainnames.html)

We remove the trailing dot simply because it is not useful (makes it annoying when you call `my_extract_result.split('.')` ) ... and in many contexts it doesn't communicate special information. It's implicit. So we strip the trailing dot: `.rstrip('.')`

I think there's no need for us to remove a space between the top-level domain (aka public suffix) and the trailing dot. It should just never be given as an input.

I think 

```
            .rstrip(".") \
            .strip()
```

is the best way.

Calling `.strip()` in the middle just seems excessive. Sorry to drag this on!

evanv:
@john-kurkowski 

I'm not sure I agree on that actually

I certainly appreciate your skepticism vis a vis supporting something like `ads.adiquity.com . `

but on the other hand, using

```
      .rstrip(".") \
      .strip()
```

you also get results like this

```
' ads.adiquity.com. '
ExtractResult(subdomain='ads.adiquity.com', domain='', suffix='')
```

Using 

```
    .strip() \
    .rstrip(".") 
```

like @john-kurkowski suggested, you end up not supporting things like `ads.adiquity.com .` and ` ads.adiquity.com . ` for example... but perhaps those shouldn't be supported in the first place, to your point.

I may be missing something but I think 

>I think there's no need for us to remove a space between the top-level domain (aka public suffix) and the trailing dot. It should just never be given as an input.

is actually an argument in favor of 

```
    .strip() \
    .rstrip(".") 
```

not 

```
      .rstrip(".") \
      .strip()
```

Although perhaps I am misunderstanding / missing something / tying myself in knots?

hangtwenty:
My goodness. You're right. I need more coffee. So yes I'm trying to favor

```
    .strip() \
    .rstrip(".") 
```

over

```
    .strip() \
    .rstrip(".") \
    .strip()
```

evanv:
either 

```
    .strip() \
    .rstrip(".") 
```

or 

```
    .strip() \
    .rstrip(".") \
    .strip()
```

make sense to me. I'm happy to update this PR either way. @john-kurkowski you leaning one way or the other right now?

hangtwenty:
@evanv yes, agreed, those are the two good options. I defer to @john-kurkowski

john-kurkowski:
My vote goes to only deleting leading/trailing whitespace. No inner whitespace.

```python
.strip() \
.rstrip(".") 
```

evanv:
Done

hangtwenty:
Great. That's my vote too. Going to merge. Thanks for your diligence here @evanv.

mauricioabreu:
Very good :)

PR #78, opened 2015-11-11
Remove unused imports

mauricioabreu:


hangtwenty:
:+1:

tldextract/tldextract.py
```diff
@@ -210,7 +208,7 @@ def __call__(self, url):
 
         if not tld and netloc and netloc[0].isdigit():
             try:
-                is_ip = socket.inet_aton(netloc)
```
john-kurkowski: Yay cleaning up unused variables! Boo this block getting more confusing what the side-effects are doing. I'll extract this into a function.

john-kurkowski:
Thanks!

tldextract/tldextract.py
john-kurkowski: Extracted in f1a83405f84cfb8c676f1485f1ecc6c64c07effa.

mauricioabreu:
You are welcome! Since I started to use this package, I am probably going to refactor some parts of the codebase. :-)

PR #80, opened 2015-11-16
Switch to py.test runner and `assert` keyword

john-kurkowski:
* Split the long test suite into a few shorter files.
* No longer distribute tests. Those are handled centrally by CI now.
* pytest-pylint plugin eliminates the need for tldextract/tests/lint.py.

py.test's automatic runner > hand-rolled test suite collection. Simple functions & `assert` keyword > JUnit-style OOP malarkey.


hangtwenty:
:beers: Good stuff

PR #82, opened 2015-12-07
Add idna 2008 support

dfeinzeig:


dfeinzeig:
@john-kurkowski , not sure what's going on with the the idna module seemingly not being installed when running tests.  any quick thoughts/pointers?

dfeinzeig:
Ok, so the issue I was having was from importing the __version__ and __doc__ before the required idna package had been installed.  My quick fix was to remove the imports from setup.py and just paste in the version and docstring.  Thoughts on how you'd really like this structured?

Also, for some reason python 2.6 and 3.5 tests pass, but not the others.  I'm at a loss, as the errors seem to be related to the test tools

john-kurkowski:
Thanks for this!

[`master`'s tests are suddenly busted too](https://travis-ci.org/john-kurkowski/tldextract/jobs/93708812). Huh. First step is to fix that. That's obscuring the issue with `tldextract.__{doc,version}__`. I can look in the next couple days.

tldextract/tests/main_test.py
```diff
@@ -36,29 +36,23 @@ def assert_extract(
         assert expected_domain == ext.domain
         assert expected_tld == ext.tld
 
-
```
john-kurkowski: This violates PEP8 no? I couldn't figure out how to enforce it with Pylint though.

tldextract/tldextract.py
```diff
@@ -39,6 +38,7 @@
 import socket
 import warnings
 
+import idna.codec
```
john-kurkowski: Why not in the alphabetical list above?

Until we get tests to run again, does this complain locally about unused import? Maybe it exports a constant you can pass to `.decode('idna')` below?

tldextract/tests/main_test.py
dfeinzeig: you mean the number of lines in between functions?  yeah, i thought it was one, but that is in between methods.  i'll make them twos.

tldextract/tldextract.py
dfeinzeig: because it is a third party import.  this is the relevant pep8:

https://www.python.org/dev/peps/pep-0008/#imports

Imports should be grouped in the following order:

standard library imports
related third party imports
local application/library specific imports
You should put a blank line between each group of imports.

no complaints from it.  consistent with the packages documented use (https://pypi.python.org/pypi/idna) and i'm currently using my branch and it works just fine.

dfeinzeig:
No problem!  This is a really great package, so I'm just glad I can contribute.

No worries about the tests.  When you dig in let me know if there is anything you'd like me to tweak in this PR.

tldextract/tldextract.py
john-kurkowski: Oh duh, I follow that ordering convention too :+1:

john-kurkowski:
Please get latest `master`. Build should be fixed.

dfeinzeig:
I back-merged your master.  All tests pass, except for this lint warning

W: 41, 0: Unused import idna.codec (unused-import)

It isn't unused though.  just doing the import idna.codec patches the python codecs.  i could use the explicit encode/decode methods but that feels a bit wrong, given that future packages could do similar types of import patching magic.  Thoughts?

dfeinzeig:
I changed the usage of idna to avoid the unused import error, even though the error is misleading.

All tests now pass.  Do I need to increment the version number in setup.py?

john-kurkowski:
I like what you came up with. Explicit > implicit import side-effect magic. (Although I see the convenience of updating the whole runtime's "idna" codec without its users changing code.)

john-kurkowski:
No need to bump the version. I'll take care of a release.

We'll need to keep digging on the `tldextract.__{doc,version}__` issues...

dfeinzeig:
Ok, cool :+1: 

The issue is that the setup.py is trying to import from tldextract.py which tries to import from the idna package, which isn't yet installed at that time, so it can't find that import.  I'm not sure how folks usually structure their setup.py to avoid copying and pasting versions and doc strings.

john-kurkowski:
We aren't the only ones who had the issue! http://stackoverflow.com/questions/17583443/what-is-the-correct-way-to-share-package-version-with-setup-py-and-the-package

Fixed on `master`. Can you rebase one last time and squash down to a few commits?

dfeinzeig:
I think this should be good to go unless you have any other comments.

john-kurkowski:
Can you squash down to 1 commit or a few commits? As is, there are almost as many commits as there are changed lines. :wink:

dfeinzeig:
Haha, sure, I'll do it later today.

dfeinzeig:
okie dokie, squashed it down to a few

john-kurkowski:
Looks like there's 1 bogus hunk in setup.py still?

You could also rebase atop `origin/master` to save the merge commits which aren't adding much to the story, but up to you.

dfeinzeig:
arggg, sorry, i don't usually do much rebasing.  i removed the extra docstring and squashed the last few commits.

john-kurkowski:
Diff looks good. I really appreciate your patience and the contribution!

dfeinzeig:
No problem!  Will you be able to release to pypi soon?  There are a few places I'd like to update to use the new version.

john-kurkowski:
Sure! Done!

dfeinzeig:
Thanks!

PR #85, opened 2015-12-23
Dont print out 10 tlds at risk of running into encoding errors

arski:
@john-kurkowski @hangtwenty here we go :)


hangtwenty:
Great thank you!

arski:
sweet! What's your policy on new pip releases? I imagine this is not enough to warrant one (though at the same time if someone runs into the bug it's kind of a deal breaker).

hangtwenty:
@john-kurkowski ?

john-kurkowski:
The policy is: on demand! :smile: Included in 1.7.4.

arski:
I like that policy - thanks a lot! :)

PR #88, opened 2016-02-05
Support gzipped PSL response

jnozsc:
Closes #87

PR #89, opened 2016-02-06
use requests instead of urllib

jnozsc:
as https://github.com/john-kurkowski/tldextract/pull/88#discussion_r52067412 mentioned

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,6 +5,7 @@ pytest==2.7.3
 pytest-gitignore==1.3
 pytest-mock==0.9.0
 pytest-pylint==0.4.0
-requests>=2.9.1
+# moniter version 3 change at https://github.com/kennethreitz/requests/blob/proposed/3.0.0/3.0-HISTORY.rst
+requests>=2.9.2,<3
 requests-file>=1.4
 tox==2.2.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,6 +6,6 @@ pytest-gitignore==1.3
 pytest-mock==0.9.0
 pytest-pylint==0.4.0
 # moniter version 3 change at https://github.com/kennethreitz/requests/blob/proposed/3.0.0/3.0-HISTORY.rst
-requests>=2.9.2,<3
+requests>=2.9.1,<3
 requests-file>=1.4
 tox==2.2.1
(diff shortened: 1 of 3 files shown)


jnozsc:
(1) `requests` can handle gzip by default http://docs.python-requests.org/en/latest/community/faq/#encoded-data
(2) `requests` can not handle `file://` so I have to add a `requests-file`

john-kurkowski:
Oh wow! I was idly wishing in [that comment](https://github.com/john-kurkowski/tldextract/pull/88#discussion_r52067412). This is pretty nice.

For tldextract 1.x, I'd like to stick to no external dependencies. Mind if I wait to merge this until 2.x?

requirements.txt
```diff
@@ -5,4 +5,6 @@ pytest==2.7.3
 pytest-gitignore==1.3
 pytest-mock==0.9.0
 pytest-pylint==0.4.0
+requests>=2.9.1
```
john-kurkowski: Should this be `>= 2.0, < 3`? Maybe I've been in Node.js too long.

john-kurkowski: Does this need to be in setup.py too?

jnozsc:
no problem, let's keep this open until 2.x

requirements.txt
john-kurkowski: Thanks!

mauricioabreu:
Good job @jnozsc 
urllib has some problems that the vendored urlblib in requests does not have.

mauricioabreu:
This PR can be closed, right? It is already merged in 2.X branch :)

john-kurkowski:
I was gonna let it autoclose when it went to `master`, but sure, closed via 6e4f9cf4cff2254bbd98ed4c9adbb744f4b5e564. Thanks @jnozsc!

john-kurkowski:
Never mind. Keeping this open until autoclose on `master`. If I close it manually, the PR is marked "Closed," not "Merged," which is misleading.

TylerLubeck:
Hi all,

Is pinning `requests>2.9.1,<3` necessary? We have a few packages that rely on both tldextract and requests 2.2 (long story), and we're getting version conflicts because of it.

I ran the test suite with requests 2.2.1, requests 2.8.1, and requests 2.9.1 and they all pass.

john-kurkowski:
The exact version range was somewhat arbitrary. I think it can be more flexible. Thanks for #98! Let's discuss there.

Issue #81, opened 2015-12-02
Use JSON instead of pickle for tld data


andresriancho:
.tld_set_snapshot and .tld_set use pickle to store the TLD information. While this is perfectly fine in most cases it brings the following issues:
 * Binary data stored in your git repository is a bad practice
 * If this library gets packaged for ubuntu/debian, the maintainer will complain. I just started using tldextract in the [w3af project](https://github.com/andresriancho/w3af/commit/72c09c57ed588caf9bc1b5c9de686b85ebe7c072) which is part of ubuntu/debian. When the package maintainer packages the next version he'll most likely dislike the binary blob
 * Pickles are "executables". A specially crafted pickle can trigger an arbitrary remote command execution when unpickled. While I did review the library for bugs and backdoors before including it in w3af, I did not read the whole pickle; which is bad for w3af user's security.

hangtwenty:
This is a good idea.

hangtwenty:
I wasn't sure if performance would degrade, looks like it would actually improve. Haha.

>  Pickle vs JSON — Which is Faster?
> If you’re here for the short answer — JSON is 25 times faster in reading (loads) and 15 times faster in writing (dumps).

[source](https://kovshenin.com/2010/pickle-vs-json-which-is-faster/)

john-kurkowski:
:+1: 

This could be the needle to finally cut a 2.0 release. If this gets done, can abandon `pickle` entirely, no need for backwards compatibility. And clean up other 1.x deprecations.

andresriancho:
Awesome, happy to see you guys liked the idea

mauricioabreu:
Good idea. :-)
I already contribute to `tldextract` project. Maybe it is a chance to contribute again.

john-kurkowski:
@mauricioabreu by all means!

I started [a 2.0 branch](https://github.com/john-kurkowski/tldextract/tree/2.0). Target any work there. Then you can run free with this issue, with less worry about backwards compatibility.

mauricioabreu:
Could not it be a plain text file?

mauricioabreu:
I mean that JSON is good for structured data, no?
We don't have a structure here, just a file separated by new lines.

Python would support this by reading lines as a normal file. What do you think @andresriancho ?

andresriancho:
I usually use JSON even for very simple data because it allows me to add more "structure" to it later (if required by the next versions of the software) without rewriting the code; but a plain text file sounds good also.

[...]

PR #91, opened 2016-03-08
Add JSON support to cached files

mauricioabreu:
Drop pickle. Addresses #81.

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -24,14 +24,11 @@ top-level domain) from the registered domain and subdomains of a URL.
 """
 
 
-try:
-    import cPickle as pickle
-except ImportError:
-    import pickle
 import collections
 from contextlib import closing
 import errno
 from functools import wraps
+import json
 import logging
 import os
 import re
@@ -109,7 +106,7 @@ class TLDExtract(object):
         Constructs a callable for extracting subdomain, domain, and suffix
         components from a URL.
 
-        Upon calling it, it first checks for a Python-pickled `cache_file`.
+        Upon calling it, it first checks for a JSON `cache_file`.
         By default, the `cache_file` will live in the tldextract directory.
 
         You can disable the caching functionality of this module  by setting `cache_file` to False.
@@ -250,21 +247,21 @@ class TLDExtract(object):
         return self._extractor
 
     def _get_cached_tld_extractor(self):
-        '''Unpickles the local TLD cache file. Returns None on IOError or other
-        unpickling error, or if this object is not set to use the cache
+        '''Read the local TLD cache file. Returns None on IOError or other
+        error, or if this object is not set to use the cache
         file.'''
         if not self.cache_file:
             return
 
         try:
-            with open(self.cache_file, 'rb') as cache_file:
+            with open(self.cache_file) as cache_file:
                 try:
-                    suffixes = pickle.load(cache_file)
-                except Exception as myriad_unpickling_errors: # pylint: disable=broad-except
+                    suffixes = frozenset(json.loads(cache_file.read()))
+                except (IOError, ValueError) as exc:
                     LOG.error(
                         "error reading TLD cache file %s: %s",
                         self.cache_file,
-                        myriad_unpickling_errors
+                        exc
                     )
                 else:
                     return _PublicSuffixListTLDExtractor(
@@ -279,7 +276,9 @@ class TLDExtract(object):
         snapshot_stream = pkg_resources.resource_stream(__name__, '.tld_set_snapshot')
         with closing(snapshot_stream) as snapshot_file:
             return _PublicSuffixListTLDExtractor(
-                self._add_extra_suffixes(pickle.load(snapshot_file))
+                self._add_extra_suffixes(
+                    frozenset(json.loads(snapshot_file.read().decode('utf-8')))
+                )
             )
 
     def _cache_tlds(self, tlds):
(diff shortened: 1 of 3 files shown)


tldextract/tldextract.py
```diff
@@ -259,12 +263,12 @@ def _get_cached_tld_extractor(self):
         try:
             with open(self.cache_file, 'rb') as cache_file:
                 try:
-                    suffixes = pickle.load(cache_file)
-                except Exception as myriad_unpickling_errors: # pylint: disable=broad-except
+                    suffixes = json.load(cache_file)
+                except Exception as exc: # pylint: disable=broad-except
```
john-kurkowski: Are `json.load`'s exceptions any easier to enumerate? Maybe `(IOError, ValueError)`?

tldextract/tldextract.py
```diff
@@ -301,7 +305,7 @@ def _cache_tlds(self, tlds):
         if self.cache_file:
             try:
                 with open(self.cache_file, 'wb') as cache_file:
-                    pickle.dump(tlds, cache_file)
+                    json.dump(tlds, cache_file, cls=SetEncoder)
```
john-kurkowski: Rather than an extra class, can this convert the frozenset to a list?

john-kurkowski:
Can you pretty print .tld_set_snapshot? Having it all on 1 line makes diffs difficult.

john-kurkowski:
Thanks for this! Just a couple minor issues. Build fail looks easy to fix too.

mauricioabreu:
Of course! :-)
I will work on it today (probably).

Thanks for your review!

mauricioabreu:
What about now @john-kurkowski ?

tldextract/tldextract.py
```diff
@@ -250,21 +247,21 @@ def _get_tld_extractor(self):
...
             return
 
         try:
-            with open(self.cache_file, 'rb') as cache_file:
+            with open(self.cache_file) as cache_file:
                 try:
-                    suffixes = pickle.load(cache_file)
-                except Exception as myriad_unpickling_errors: # pylint: disable=broad-except
+                    suffixes = json.loads(cache_file.read())
+                except (IOError, ValueError) as exc: # pylint: disable=broad-except
```
john-kurkowski: Sweet. Probably don't need to disable broad-except then.

tldextract/tldextract.py
```diff
@@ -229,7 +226,7 @@ def _get_tld_extractor(self):
         if self._extractor:
             return self._extractor
 
-        tlds = frozenset()
```
john-kurkowski: Oh, I wasn't suggesting to ditch `frozenset`. We [still need its performance](https://github.com/john-kurkowski/tldextract/blob/038217b79f378da8b055e30d65a1b83e29926514/tldextract/tldextract.py#L352). I meant, coerce from `frozenset` to something JSON serializable, like `list`, _when_ you need to serialize.

mauricioabreu: You are right! Since we changed the way JSON is being encoded/decoded we can use frozenset again.

tldextract/tldextract.py
```diff
@@ -334,7 +331,7 @@ def get_tlds_from_raw_suffix_list_data(suffix_list_source, include_psl_private_d
         text, _, _ = suffix_list_source.partition('// ===BEGIN PRIVATE DOMAINS===')
 
     tld_iter = (m.group('suffix') for m in PUBLIC_SUFFIX_RE.finditer(text))
-    return frozenset(tld_iter)
+    return list(tld_iter)
```
john-kurkowski: Other code paths besides JSON serializing use this, and will get the wrong type. Try coercing to a `list` only when serializing, e.g.

```python
json.dump(list(my_frozenset_of_tlds), some_file)
```

tldextract/tldextract.py
```diff
@@ -250,21 +247,21 @@ def _get_tld_extractor(self):
...
+        error, or if this object is not set to use the cache
         file.'''
         if not self.cache_file:
             return
 
         try:
-            with open(self.cache_file, 'rb') as cache_file:
+            with open(self.cache_file) as cache_file:
                 try:
-                    suffixes = pickle.load(cache_file)
```
john-kurkowski: If you serialize a `list` to JSON, anything that was `pickle.load` before will have to convert it back to `frozenset`.

mauricioabreu: I don't think I get it. Can you explain?

tldextract/tldextract.py
mauricioabreu: Right! Going to change it.

tldextract/tldextract.py
john-kurkowski: Sure.

```python
>>> import json
>>> foo = frozenset([1,2,3])
>>> f = open('bar.json', 'w')
>>> json.dump(foo, f)
TypeError: frozenset([1, 2, 3]) is not JSON serializable
>>> json.dump(list(foo), f)
>>> assert type(foo) == frozenset # okay!
>>> f.close()
>>> baz = json.load(open('bar.json'))
>>> assert type(baz) == frozenset
AssertionError
```

mauricioabreu: Ah, you are talking about the comment above, right? It will be solved by `only` coercing when serializing I guess so. Am I missing something?

john-kurkowski: I mean you must serialize from a `frozenset` to JSON and also deserialize from JSON back to  a `frozenset`.

john-kurkowski:
Latest diff LGTM! :+1: :sparkles: Thanks for sticking with it!

tldextract/tldextract.py
mauricioabreu: I changed the code again. Sorry, it is like 03:30 AM here in Brazil :)

mauricioabreu:
I learned a lot with this PR. You are a very good person. Thank you!

PR #92, opened 2016-03-13
2.0

john-kurkowski:
This is the long-running branch for 2.0. Changes are already enumerated in CHANGELOG.md. Mostly cruft cleanup, no new features. No other changes planned.

If this looks good, let's merge to master and publish 2.0rc1.

PR #93, opened 2016-03-24
Fix incorrect StringIO usage for Python 3

Randati:
```python
buf = StringIO(response.read())
```
was failing with
```
TypeError: initial_value must be str or None, not bytes
```


yprez:
+1

Also, if you do `from StringIO import StringIO as BytesIO` or `from io import BytesIO as StringIO` you won't need the 2nd try-except.

john-kurkowski:
This problem will go away in the next planned release, 2.0 (#92). No "io" imports any more.

john-kurkowski:
With all "io" references now removed, this shouldn't be necessary any more. Thank you though!

PR #94, opened 2016-03-24
Fix TypeError: initial_value must be str or None, not bytes on python3

cansin:
We were having problems using tldextract when local `.tld_set_snapshot` cache is missing. We realized it is related with a str vs. byte problem on python3. tox passed for python2.7 and python3.4.

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -32,6 +32,7 @@ import collections
 from contextlib import closing
 import errno
 from functools import wraps
+from io import BytesIO
 import logging
 import os
 import re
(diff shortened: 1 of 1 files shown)


tldextract/tldextract.py
```diff
@@ -68,12 +68,12 @@ def resource_stream(cls, _, resource_name):
     # Python 2
     from urllib2 import urlopen
     from urlparse import scheme_chars
-    from StringIO import StringIO
+    from io import BytesIO
```
yprez: If this works for all versions, why keep it in the try-except?

john-kurkowski:
Does this subsume #93?

john-kurkowski:
This problem will go away in the next planned release, 2.0 (#92). No "io" imports any more.

tldextract/tldextract.py
cansin: @yprez you're right, updated.

cansin:
@john-kurkowski yes it does.

cansin:
Feel free to favor https://github.com/john-kurkowski/tldextract/pull/93 over this if you think it is safer to not to touch python 2.x derivatives.

john-kurkowski:
With all "io" references now removed, this shouldn't be necessary any more. Thank you though!

PR #95, opened 2016-04-01
convert bytes to str for StringIO

anneFly:
In Python3 the result of HTTPResponse.read() is of type bytes. StringIO cannot
handle this, it will throw the error: TypeError: initial_value must be str or
None, not bytes. This commit fixes this behaviour.


anneFly:
oh sorry, didn't see that there was already a PR. I close this one.

PR #97, opened 2016-04-21
Include tests in release source tarball

EdwardBetts:
I'm thinking about packaging tldextract so it can be included in Debian. It would be great to include the test suite in the pypi tarball so the tests can be run when the Debian package is built.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,1 +1,2 @@
 include tldextract/.tld_set_snapshot
+recursive-include tests *.py *.dat


john-kurkowski:
Sure! Thanks!

In dedcc127fdccbcc2a478a79fb52636317961b674, I removed the tests from the dist, thinking, "TravisCI represents all test environments everywhere!" This was misguided. If you wanna run the tests yourself, people should be able.

PR #104, opened 2016-08-01
Update tests

medecau:
This pull-request:

- Drops test support for python 2.6.
  - too many errors related to pylint.
  - decided to keep pylint and drop python 2.6.
- Adds requests version variations to tox config similar to travis.
  - `tox -l` for a list of all environments included
  - `detox` to run tests in parallel
- Moves test dependencies into tox.
- Adds a new environment for code style conformance validation.
  - `pycodestyle`
  - fixed some code style errors
- Changes travis configuration to use tox.
- Changes travis configuration to use containers.


john-kurkowski:
Thank you for your contribution. Can I ask, what prompted this? How is this better?

.travis.yml
```diff
@@ -1,16 +1,27 @@
+sudo: false
```
john-kurkowski: This is Travis's default

tox.ini
```diff
@@ -1,6 +1,22 @@
 [tox]
-envlist = py26, py27, py33, py34, py35, pypy
+envlist = py{27,33,34,35,py}-requests{,2.1.0,2.8.1,2.9.1},codestyle
 
 [testenv]
+deps =
+    pylint
```
john-kurkowski: What's the advantage of duplicating these developer dependencies here? They seemed just fine in requirements.txt?

tox.ini
```diff
@@ -1,6 +1,22 @@
...
+    py{26,27,33,34,35,py}-requests: requests
+    py{26,27,33,34,35,py}-requests2.1.0: requests==2.1.0
+    py{26,27,33,34,35,py}-requests2.8.1: requests==2.8.1
+    py{26,27,33,34,35,py}-requests2.9.1: requests==2.9.1
+
 commands = py.test
-deps = -rrequirements.txt
+
+[testenv:codestyle]
+deps = pycodestyle
```
john-kurkowski: What's the benefit of pycodestyle _as well as_ pylint?

.travis.yml
```diff
@@ -1,16 +1,27 @@
...
-  - "2.7"
-  - "3.3"
-  - "3.4"
-  - "3.5"
-install:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.6* ]]; then export PYTEST_ADDOPTS='-p no:pylint'; fi
-  - pip install -r requirements.txt
-  - pip install --upgrade --force-reinstall requests==$REQUESTS_VERSION
-script: py.test
+    - TOXENV=py27-requests
```
john-kurkowski: This seems to have gotten longer in terms of lines & execution time. Compare the last Travis build, [build#120](https://travis-ci.org/john-kurkowski/tldextract/builds/142037274), with this PR's, [build#121](https://travis-ci.org/john-kurkowski/tldextract/builds/148828810).

It's also slightly less clear which Python version Travis is running. At first glance, it looks like it's testing Python 3.5 over and over.

What's the reason for this change?

tox.ini
medecau: Good question. I don't know much about pylint to be able to provide a complete answer though.
In my experience I find that pylint and pycodestyle can often find different things.
While pylint often lists harder to find errors like bad variable names pycodestyle focuses on PEP-8 compliance.

tox.ini
medecau: This way test dependencies are defined in the same file that defines other test environment options.
Both tests and package install no longer use requirements.txt

.travis.yml
medecau: The time taken to run all tests is directly correlated to the number of environments defined.

In this pull-request I chose to set all the possible tox environment variations. It is up to the maintainers of the project to decide exactly which environments to use. I can definitely reduce this to a more manageable set of environments.

I agree it is less clear identify which test version failed. The alternative is that the test environment definition is kept in only one file, the tox config file. And will be identical for both local and remote test runs.

.travis.yml
medecau: True.
According to the docs this is the default for all repos activated since 2015. And was already the default behaviour for tldextract.

Should I remove it?

john-kurkowski: Eh, you're right, explicit > implicit

tox.ini
john-kurkowski: PEP-8 enforcement, neat. I've been looking for that.

medecau:
> What prompted this?

There wasn’t a single cause for the changes or I can no longer remember what it was.
But I made the changes. I think they are good and stand by them.

> How is this better?

Before remote tests were more exhaustive than local tests. i.e.: local tests could pass while remote tests failed. With these changes the steps for local tests are the same as for remote tests. 
Devs are able to test on all relevant environments before pushing to remote. And during development it is possible to run only one set of tests for the specific environment the changes affect.
When Python 3.6 is released it will be added to `tox.ini` along with one line to `.travis.yml`.
There is an environment solely for code style. In the future you can add environments for doc style and security.

> Do developers need to run all environment permutations all the time?

No. Developers can run tests against only one environment or a small set of environments. e.g.: `tox -e py35-requests-2.1.0`
[More info on tox docs.](https://tox.readthedocs.io/en/latest/)

> Is Tox the new hotness?

Probably. Depends on the definition of ‘new’ and ‘hotness’. I’ll have to leave this one up to you.
Here are a few commits of tox being added to some influential projects:

- [2016 django](https://github.com/django/django/commit/09d38746bafa4b3dd9fc21ffab822ed53087d794#diff-b91f3d5bd63fcd17221b267e851608e8)
- [2015 ansible](https://github.com/ansible/ansible/commit/3d67e9e0c0df18e5c82e62fdb79820724dbe2577#diff-b91f3d5bd63fcd17221b267e851608e8)
- [2012 httpie](https://github.com/jkbrzt/httpie/commit/0d129d5f692d76ad34e7b1d0f9241eea57ba02d1#diff-b91f3d5bd63fcd17221b267e851608e8)
- [2010 flask](https://github.com/pallets/flask/commit/c18f032a822fa6db560c440d7faa7c615826aa30#diff-b91f3d5bd63fcd17221b267e851608e8)

medecau:
After posting the previous comment I noticed it included an appeal to authority.

Let me try with my personal experience. Anecdotal as it may be.

I've been using tox for about the same time as I've been using py.test. A few years now. I don’t consider it *new*. I find it extremely usefull. Even if it may be *hot* I find it to be usefully *hot*. Using tox on a project gives me the guarantee that whatever changes I make do not brake the project at *any* level that tox is able to test. That is, the project builds, it installs, dependencies do not brake, and tests pass. I have not explored running the test on different OSs. For now I leave that to travis-ci.

I was also going to list a bunch of tox features and guarantees that tox provides but decided against it because the tox wiki does a much better job than I could possibly do. See [tox docs](https://tox.readthedocs.io/en/latest/).

I also remembered that a similar clarification was requested on a different project. See [transitions issue #123](https://github.com/tyarkoni/transitions/pull/123#issuecomment-234644793)

.travis.yml
```diff
@@ -1,16 +1,15 @@
...
-  - "2.7"
-  - "3.3"
-  - "3.4"
-  - "3.5"
-install:
-  - if [[ $TRAVIS_PYTHON_VERSION == 2.6* ]]; then export PYTEST_ADDOPTS='-p no:pylint'; fi
-  - pip install -r requirements.txt
-  - pip install --upgrade --force-reinstall requests==$REQUESTS_VERSION
-script: py.test
+    - TOXENV=py27-requests-current
```
john-kurkowski: Minor: does Travis have any syntax for syncing this list of envs to run with tox.ini? Too bad it has to be in 2 places. But it's not too bad.

john-kurkowski:
Thank you very much for the explanation. I like seeing the other examples and discussions. From this PR's description alone, I wasn't sure what to make of the increased reliance on Tox.

> Before remote tests were more exhaustive than local tests.

I hadn't realized that. The way this PR smooths that over is nice.

And no argument against PEP-8 compliance from me. I was looking for that anyway.

So, this new direction sounds good. I'll bite.

john-kurkowski:
I'm trying to check out your branch locally, and try the new test suite, as if I were a newb. [The README's Contribute section](https://github.com/john-kurkowski/tldextract/tree/2.0.1#contribute) doesn't seem to apply any more. There's no requirements.txt. I get `tox: command not found`. Can you update README.md?

Hopefully that section can remain as succinct as it is. It doesn't need to be nearly as detailed as [Django's addition of tox](https://github.com/django/django/commit/09d38746bafa4b3dd9fc21ffab822ed53087d794#diff-5830044babd91b5beb690dd0104b4d56R48).

.travis.yml
medecau: You could run tox without any TOXENV vars but that would collapse all the environments into a single one. Tox would then run all permutations but it would be harder to which tests failed from the travis interface.

By defining the different environments it forces travis to display the results for the different environments.

medecau: What the above means in usability is that currently if a test fails in one of the environments travis will display the failed environment from the UI. e.g.: https://travis-ci.org/john-kurkowski/tldextract/builds/93708810

If no TOXENV vars are defined then tox will run all permutations under a single build, hiding the details of which environment failed deep in the logs. Tox makes it easy to see which envs failed but not as nice as travis web UI.

medecau:
You may have received a notification for an update on this PR. Please ignore that.

I went back to the README and am trying to figure out the steps for contribution.
The current steps for installation are not the steps for contribution.

For contribution I would simply mention pull-requests as the way to go. And request that contributors validate that all tests pass before publishing.

john-kurkowski:
Oh, I'm not asking you to add more info to that section than what's already there. Just to update the incorrect info.

In particular, I'm trying to run your PR locally, but the instructions in that section don't work any more.

john-kurkowski:
Whew, very simple instructions. Tried it out locally. Works great. Thank you for seeing this through!

medecau:
You are welcome!

Thanks for taking the time to review the pull-request.

PR #126, opened 2017-04-28
Feature property fqdn

jvanasco:
this is a suggestion for a convenience function

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -131,7 +131,8 @@ class ExtractResult(collections.namedtuple('ExtractResult', 'subdomain domain su
         ''
         """
         if self.domain and self.suffix:
-            return '.'.join([i for i in [self.subdomain, self.domain, self.suffix] if i])
+            # self is the namedtuple (subdomain domain suffix)
+            return '.'.join([i for i in self if i])
         return ''
 
 
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -123,12 +123,16 @@ class ExtractResult(collections.namedtuple('ExtractResult', 'subdomain domain su
     @property
     def fqdn(self):
         """
-        Returns a Fully Qualified Domain Name.
+        Returns a Fully Qualified Domain Name, if there is a proper domain/suffix.
 
         >>> extract('http://forums.bbc.co.uk/path/to/file').fqdn
         'forums.bbc.co.uk'
+        >>> extract('http://localhost:8080').fqdn
+        ''
         """
-        return '.'.join([i for i in [self.subdomain, self.domain, self.suffix] if i])
+        if self.domain and self.suffix:
+            return '.'.join([i for i in [self.subdomain, self.domain, self.suffix] if i])
+        return ''
 
 
 class TLDExtract(object):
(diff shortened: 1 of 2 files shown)


tldextract/tldextract.py
```diff
@@ -120,6 +120,20 @@ def registered_domain(self):
...
+        """
+        Returns a Fully Qualified Domain Name, if there is a proper domain/suffix.
+
+        >>> extract('http://forums.bbc.co.uk/path/to/file').fqdn
+        'forums.bbc.co.uk'
+        >>> extract('http://localhost:8080').fqdn
+        ''
+        """
+        if self.domain and self.suffix:
+            return '.'.join([i for i in [self.subdomain, self.domain, self.suffix] if i])
```
john-kurkowski: Can save 2 lists with `'.'.join(i for i in self if i)`.

john-kurkowski:
Rebase atop latest master. That should fix Travis.

jvanasco:
can you share any info on how you run/setup the tox tests under python 2.7 ?

I tried writing tests and running tox as described in the docs -- and it's collecting 2k+ tests and jamming my box.

looking at the travis build for this, it looks like the python 2.7 tests may be running under python 3 too.  i won't have time to investigate more until the weekend, so just wanted to bring it up here in case you know of something that I'm missing.

john-kurkowski:
`tox -e <one-env-here>`, like [this line of the README](https://github.com/john-kurkowski/tldextract/blame/2.0.2/README.md#L222), runs the test suite for only 1 Python version. That works well for me for local development. That'll give you a much faster feedback loop. It's not bulletproof. That's okay. Once you push your work, Travis will run the entire test suite matrix.

jvanasco:
That's the command that is collecting 2k+ tests for me with Python 2.7 -- with or without a fresh virtual env.

That didn't happen under python 3 though, which is what puzzles me.

john-kurkowski:
Hmm. I'm not seeing that.

```zsh
$ python --version                                                                                                                        
Python 2.7.13
 ~/S…/tldextract $ tox -e py27-requests-current -- --collect-only 2>&1  | grep collected                                                                   
collected 46 items
 ~/S…/tldextract $ tox -e py36-requests-current -- --collect-only 2>&1  | grep collected                                                                   
collected 34 items
```

jvanasco:
thanks for humoring me on this.  tox seems to be broken on this build of OSX.  i'm running 2.7 tests on an ubuntu box fine now.

tldextract/tldextract.py
john-kurkowski: bump

jvanasco: still working on tests.

i will gladly add this change, but i thought being explicit would be better for reading/maintenance.

jvanasco:
Ok. changes made and tests added.

Sorry for the travis overload - I did most of my work in a different branch and merged back to avoid issues... then had to reorganize the tests to make them pass.

The test reorganization is pretty simple.  Tox doesn't want to see more than 5 args per function.  instead of changing the rule, I slightly altered the way `assert_extract` works.  Now the URL is first (instead of last) and the "expected" elements are a tuple.  This makes reading the tests much easier.

john-kurkowski:
Wow! So exhaustive! I'd have been fine with the doctest but this is great. 😋  Thank you!

Issue #122, opened 2017-04-25
python 2.7 is often 'narrow', doesn't catch unicode error


jvanasco:
this example will work on python3.x but breaks on many default python2.7 installs (including many OSX and Ubuntu):

    import tldextract
    tldextract.extract('xn--vi8hiv.ws')

it will raise:

    ValueError: unichr() arg not in range(0x10000) (narrow Python build)

The problem is in that this block only checks for a `UnicodeError`:  https://github.com/john-kurkowski/tldextract/blob/2.0.2/tldextract/tldextract.py#L198-L204

however if Python wasn't compiled with support for 'wide' unicode, this will raise a `ValueError`

there are many strategies for supporting 'narrow' unicode, or this could just be caught and skipped.

john-kurkowski:
> there are many strategies for supporting 'narrow' unicode, or this could just be caught and skipped.

I'm fine with "caught and skipped." Though I'm curious what the strategies and their tradeoffs are.

jvanasco:
the two most popular methods on stackoverflow/github/maillinglists I've found are:

use unicode-escape to store as a utf16 pair
* http://stackoverflow.com/questions/7105874/valueerror-unichr-arg-not-in-range0x10000-narrow-python-build-please-hel

rebuild it using struct
* https://github.com/euske/pdfminer/pull/148/commits/cea7eefde6301e377bb4a3a499fb2e2ace8c2a78

I saw a few other approaches.  I haven't had time to dig into them yet.

I've had lots of issues with unicode over the years, but it's been a long time since the last one.  this one surprised me!

john-kurkowski:
Whew, that's not bad. I was dreading new 3rd party packages or files.

jvanasco:
Adding for reference: PEP 261 - support for wide characters

https://www.python.org/dev/peps/pep-0261/

hangtwenty:
It's 2020 and Python 2.x is dead.

PR #130, opened 2017-05-26
Fix narrow python

jvanasco:
This is a lightweight attempt to handle the "narrow Python build" I mentioned in #122

If a ValueError is detected, it looks in the exception message (Py2 and Py3 compliant) to see if the 'narrow Python build' warning is detected.  If so, it emits the message as a `warning` and returns the label unaltered.

I only put the docstring on `decode_punycode` to pass tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 After upgrading, update your cache file by deleting it or via `tldextract
 --update`.
 
+## 2.1.1 (?)
+
+* Bugfixes
+    * Basic support for 'narrow' Python2 builds ([#122](https://github.com/john-kurkowski/tldextract/issues/122))
+
 ## 2.1.0 (2017-05-24)
 
 * Features
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -216,3 +216,19 @@ def test_result_as_dict():
                      'domain': 'google',
                      'suffix': 'com'}
     assert result._asdict() == expected_dict
+
+
+def test_punycode_extended():
+    """
+    "narrow python" builds may break on certain unicode characters
+    see https://github.com/john-kurkowski/tldextract/issues/122
+    """
+    # this one breaks on narrow python, needs the bypass
+    assert_extract('xn--vi8hiv.ws',
+                   ('xn--vi8hiv.ws', '', 'xn--vi8hiv', 'ws'))
+
+    # these two are the same domain, different representations
+    assert_extract(u'http://➡.ws/♥',
+                   (u'➡.ws', '', u'\u27a1', 'ws'))
+    assert_extract('http://\xe2\x9e\xa1.ws/\xe2\x99\xa5',
+                   ('\xe2\x9e\xa1.ws', '', '\xe2\x9e\xa1', 'ws'))
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -124,11 +124,11 @@ def test_puny_with_non_puny():
 def test_punycode_narrow_python():
     """
     see https://github.com/john-kurkowski/tldextract/issues/122
-    
+
     Python2 through 3.3 stores unicode data as either UCS-2 (roughly UTF-16)
     or UCS-4 (UTF-32). The default storage format was UCS-2; supporting UCS-4
     requires compiling Python with extended support using this configure option:
-    
+
         ./configure --enable-unicode=ucs4
 
     The max unicode point is available by the variable `sys.maxunicode`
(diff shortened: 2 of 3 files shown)


tests/main_test.py
```diff
@@ -216,3 +216,19 @@ def test_result_as_dict():
                      'domain': 'google',
                      'suffix': 'com'}
     assert result._asdict() == expected_dict
+
+
+def test_punycode_extended():
+    """
+    "narrow python" builds may break on certain unicode characters
```
john-kurkowski: Can you colocate this test with other punycode test(s)? Can you expand this comment how it differs from those other tests? 

CI passes with or without the library code change in this PR. Worth mentioning that? And what the test _should_ do on a narrow build? I'm fine with "should." For this edge case, I don't care to maintain a narrow build environment in CI. :blush:

tldextract/tldextract.py
```diff
@@ -227,11 +229,19 @@ def __call__(self, url):
         labels = netloc.split(".")
 
         def decode_punycode(label):
+            """helper function"""
```
john-kurkowski: While you're in here, could be a little more specific?

CHANGELOG.md
```diff
@@ -3,6 +3,12 @@
 After upgrading, update your cache file by deleting it or via `tldextract
 --update`.
 
+## 2.1.0 (2017-05-24)
```
john-kurkowski: Accidental commit slip in here?

jvanasco: I'll try to rebase again before fixing the other issues; might an issue with this version of git. If so, I'll try from another machine.  Github keeps thinking the existing merge already in master is new.

jvanasco:
I largely addressed the changes.  I'll repatch against a new master if they're okay - that should get past the git issues.

I'm possibly doing a bad job explaining the tests/passing, so I'll try rephrasing it a bit here...

If one has a "narrow" Python build (which is the default for python through 3.3), then the punycode decoding will raise an exception on a UTF-16 character.  This fix just leaves it as punycode - which is what happens if a unicode error is encountered.  I don't think it would be likely to submit an encoded UTF-16 into the function - though it may be possible.  (the domain that triggers this, incidentally, is a url shortener that turns links into emojis - because of course)

Tests could easily be set to only run on narrow builds... I thought it would make sense to run in all environments

    import sys
    import pytest


    def possible_narrow_unicode():
        """
        Python3 doesn't have a "narrow" unicode problem.
        Python2 may either be built with a narrow unicode, or have expanded support
        """
        if sys.version_info < (3, 0):
            if sys.maxunicode <= 65535:
                return True
        return False


    SKIP_TESTS_NARROW_UNICODE = pytest.mark.skipif(not possible_narrow_unicode(),
                                                   reason="not a narrow unicode build")

    @SKIP_TESTS_NARROW_UNICODE
    def test_punycode_narrow_python():
        ....

john-kurkowski:
That sample `skipif` code is nice. I'd almost prefer it, to document the test better: the test only applies in situation X. Unfortunately, I don't think the test suite exercises situation X. Not without further changes to tox.ini and .travis.yml. So, better a test that always runs, vs. a test that never runs and is forgotten.

Your expanded test docstring is what I was looking for. Can it also clarify the new test case is redundant on a wide build, that it is primarily to cover a narrow build? Again, your `skipif` code would've showed this rather than tell it. But no need for further maintenance hoops.

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
 After upgrading, update your cache file by deleting it or via `tldextract
 --update`.
 
+## 2.1.1 (?)
```
john-kurkowski: No need to fill out release details in this PR. I can take care of that when I release.

john-kurkowski:
> I'll repatch against a new master if they're okay - that should get past the git issues.

Rebasing atop the latest upstream branch, Git would've seen your original 0eeafb5e224f77938e11d9a2a754c9f22ee7bd26 was an empty commit and omitted it. Or interactively, stopped to suggest omitting it. Now that this PR contains an actual CHANGELOG.md diff, rebasing will update the diff to contain just that actual, new change. No duplicate diff with `master`. (However, [I've asked here](https://github.com/john-kurkowski/tldextract/pull/130#discussion_r118845892), can you delete the commit entirely?)

jvanasco:
I would have preferred for the skipif, and decided against it, for the same reasons as you. I looked into supporting a narrow Python build and it just didn't seem worth setting up a custom Python binary.  I have narrow and wide builds on my boxes, most people don't.

I'll give the rebase another try - it just wasn't coming out cleanly .  If it doesn't work, I diff'd this branch into a new one with a proper, clean, single-commit. I'll replace this branch with that one if I can't git to play nice, as I don't have time to upgrade it (every box here has an older git)

hangtwenty:
@jvanasco do you mind resolving the conflict? thanks!

PR #132, opened 2017-07-19
Returns the registered domain in lowercase

bastbnl:
Thanks for building and maintaining this incredibly useful module.

I added a small update to `registered_domain`, returning the value in lowercase. I ran into a potential issue when comparing user provided input provided. I verified this PR on Python 2.7

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -110,15 +110,17 @@ class ExtractResult(collections.namedtuple('ExtractResult', 'subdomain domain su
     @property
     def registered_domain(self):
         """
-        Joins the domain and suffix fields with a dot, if they're both set.
+        Creates a new string Joins the domain and suffix fields with a dot, if they're both set.
 
         >>> extract('http://forums.bbc.co.uk').registered_domain
         'bbc.co.uk'
         >>> extract('http://localhost:8080').registered_domain
         ''
+        >>> extract('http://cdn.dev.BrwnPpr.com').registered_domain
+        'brwnppr.com'
         """
         if self.domain and self.suffix:
-            return self.domain + '.' + self.suffix
+            return '{domain}.{suffix}'.format(domain=self.domain, suffix=self.suffix, ).lower()
         return ''
 
     @property
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -110,7 +110,7 @@ class ExtractResult(collections.namedtuple('ExtractResult', 'subdomain domain su
     @property
     def registered_domain(self):
         """
-        Creates a new string, containing the the domain (in lowercase), a dot and 
+        Creates a new string, containing the the domain (in lowercase), a dot and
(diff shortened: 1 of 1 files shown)


john-kurkowski:
Thank you for contributing!

I'm not sure I want this. It'd be a breaking change. Leaving user input untouched is intentional. Without this PR, the user can map back to their original string. With this PR, that mapping is lost. If lowercase domain names are key to the user, they can safely call `.lower()` themselves. That sounds more flexible than this lib forcing lowercase on everyone.

bastbnl:
Hi,

I can totally understand that thought and I'd share it if I hadn't caught myself adding `.lower()` in both of the projects I used tldextract in. 

I'll update the pr and add a keyword argument, enabling the user to normalize the result. Will that work?

Another reason I added it, btw, is that DNS is case insensitive, so returning it in lowercase should not cause any serious errors.

john-kurkowski:
Maybe I don't understand your "potential issue when comparing user provided input provided." Can you explain further? Is this a common issue? I only see an 8 character shortcut for 1 property. A shortcut which is safe to take in your use case but loses information in other use cases.

> I'll update the pr and add a keyword argument

What will that look like? Doesn't maintaining it as a keyword arg make it more characters to call?

```python
# before
result.registered_domain.lower()

# after
result.registered_domain(lower=True)
```

At that point, callers might as well use the familiar, shorter `.lower()`? For backwards compat, I also worry about changing from property to method.

bastbnl:
Here are my use-cases:
* First is an app that enables the user access to an URL. We check to see if this root domain is on the list prior to adding it. We don't have to do anything when you add `www.brwnppr.com` while `brwnppr.com` is on that list. We're working with user input here, so we want to match `BrwnPpr.com` too.
* Next is an app that automatically generates short URI unless the link added by the user points to the local domain. So we don't create a short URI when we find a URI like `http://docs.brwnppr.com/static/xls/contacts.xls` in the user generated content when the organization website is at www.brwnppr.com. The idea is that we prevent unintended leaking of product and/or path info.

Does that make sense?

I wrote the comment concerning the `lower=True` option with a function in mind. I forgot it was a property. Sorry about that.

We could go for `registered_domain_ex`, but that doesn't look too good.

john-kurkowski:
Thank you for the detail!

Case-insensitive comparisons of user input are a common thing to do, sure. This library could've been built from the ground up to facilitate that, discarding the original input's case. But given its history, preserving case, it can't assume your use case (and only in 1 property). I think the most common remedy is your case is well known and compact: `str.lower` all the things you're comparing.

> We could go for `registered_domain_ex`, but that doesn't look too good.

Right. Another property increases API surface area. It's gotta be compelling to do so. Any suffix to distinguish the lowercased version from the preserved-case version will either be unclear, or more verbose than calling `str.lower` yourself.

So, history + familiar remedy + short remedy, I'll decline this.

PR #137, opened 2017-08-17
Fix __init__.py when pkg_resources is missing

calebmadrigal:
Simply copied the `pkg_resources` fake fallback from `tldextract.py` into `__init__.py` to prevent failure if `pkg_resources` is missing.

### DIFF
--- a/tldextract/__init__.py
+++ b/tldextract/__init__.py
@@ -1,13 +1,4 @@
 """Export tldextract's public interface."""
 
-import os
-
-try:
-    import pkg_resources
-    __version__ = pkg_resources.get_distribution('tldextract').version  # pylint: disable=no-member
-except ImportError:
-    __version__ = '(local)'
-except pkg_resources.DistributionNotFound:
-    __version__ = '(local)'
-
+from .cli import __version__
 from .tldextract import extract, TLDExtract
--- a/tldextract/__init__.py
+++ b/tldextract/__init__.py
@@ -11,4 +11,3 @@ except pkg_resources.DistributionNotFound:
     __version__ = '(local)'
 
 from .tldextract import extract, TLDExtract
-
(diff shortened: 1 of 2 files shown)


tldextract/__init__.py
```diff
@@ -1,6 +1,20 @@
...
+try:
+    import pkg_resources
+except ImportError:
+    class pkg_resources(object):  # pylint: disable=invalid-name
+
+        """Fake pkg_resources interface which falls back to getting resources
+        inside `tldextract`'s directory.
+        """
+        @classmethod
+        def resource_stream(cls, _, resource_name):
```
john-kurkowski: In this file, the fake fallback is only used for `pkg_resources.get_distribution`, a few lines down (outside of the range I can comment on). This fake doesn't define that method. Does this trade an `ImportError` for an `AttributeError`?

john-kurkowski:
Surprised this has slipped past for so long! Any thoughts how to add test coverage?

tldextract/__init__.py
calebmadrigal: Nice catch. We would need to define `get_distribution`. I don't know if there's a good way to get the version of the package without `pkg_resources`. Perhaps we could just add `ImportError` to the exception handler that catches `pkg_resources.DistributionNotFound` below?

calebmadrigal:
@john-kurkowski What do you think of this alternative approach?

john-kurkowski:
I like it! Just need to fix or work around the CI failure.

john-kurkowski:
I deduped the `__version__` calculation. Not sure why we didn't do that before. Does that work?

calebmadrigal:
Looks good to me!

john-kurkowski:
🙌

PR #139, opened 2017-10-06
Allow setting timeouts for public suffix list downloads

TylerLubeck:
Heya!

We've been working with some apps that use tldextract to initialize some things at startup time. We recently put these apps in a datacenter that doesn't have much access to the outside world. Then when tldextract tries to reach out for refreshed publix suffix lists, the apps hang indefinitely since the requests never go through.

We'll probably end up solving our specific problem by supplying a cache file and setting `TLDEXTRACT_CACHE` per the docs, but it seems like the kind of problem where being able to set a timeout would be beneficial.

Would you be interested in pursuing something like this?

Thanks!

### DIFF
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -1,9 +1,13 @@
 # -*- coding: utf-8 -*-
 '''Main tldextract unit tests.'''
 
+import sys
+
 import responses
 import tldextract
 from .helpers import temporary_file
+if sys.version_info >= (3,):  # pragma: no cover
+    unicode = str  # pylint: disable=invalid-name,redefined-builtin
 
 
 # pylint: disable=invalid-name
@@ -219,11 +223,11 @@ def test_result_as_dict():
     assert result._asdict() == expected_dict
 
 
-@responses.activate # pylint: disable=no-member
+@responses.activate  # pylint: disable=no-member
 def test_cache_timeouts():
     server = 'http://some-server.com'
-    responses.add( # pylint: disable=no-member
-        responses.GET, # pylint: disable=no-member
+    responses.add(  # pylint: disable=no-member
+        responses.GET,  # pylint: disable=no-member
         server,
         status=408
     )
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 '''Main tldextract unit tests.'''
 
+import responses
 import tldextract
 from .helpers import temporary_file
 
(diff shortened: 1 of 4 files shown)


tests/main_test.py
```diff
@@ -216,3 +217,15 @@ def test_result_as_dict():
                      'domain': 'google',
                      'suffix': 'com'}
     assert result._asdict() == expected_dict
+
+
+@responses.activate # pylint: disable=no-member
+def test_cache_timeouts():
+    server = 'http://some-server.com'
+    responses.add( # pylint: disable=no-member
+        responses.GET, # pylint: disable=no-member
```
TylerLubeck: the test suite complained about `no-member`, and I'm not totally sure how to fix that properly. I saw this in a few other places in code, but if you can point me in the direction of how to do it right then I'd be happy to do so!

tldextract/tldextract.py
```diff
@@ -204,6 +218,19 @@ def __init__(self, cache_file=CACHE_FILE, suffix_list_urls=PUBLIC_SUFFIX_LIST_UR
         self.extra_suffixes = extra_suffixes
         self._extractor = None
 
+        self.cache_fetch_timeout = None
+        if CACHE_TIMEOUT is not None:
```
sseg: This seems to take the opposite precedence order from what CACHE_FILE takes: with the `cache_file` argument you can override the environment variable using the keyword argument. On the other hand, if there exists a valid value for `cache_fetch_timeout` available from the environment then the API does not allow overriding it with the keyword argument.

Is there an advantage to preferring this argument precedence here?

TylerLubeck: Can I blame coding after bed time? I'm gonna blame coding after bed time.

Good catch as always @sseg, thanks!

TylerLubeck:
Heya @john-kurkowski - any thoughts on this?

tests/main_test.py
john-kurkowski: 🤷‍♂️ This seems fine. The alternative is [disabling linting of `responses` completely](https://github.com/getsentry/responses/issues/74#issuecomment-158025086).

tldextract/tldextract.py
```diff
@@ -204,6 +218,18 @@ def __init__(self, cache_file=CACHE_FILE, suffix_list_urls=PUBLIC_SUFFIX_LIST_UR
         self.extra_suffixes = extra_suffixes
         self._extractor = None
 
+        self.cache_fetch_timeout = cache_fetch_timeout
+        if not isinstance(self.cache_fetch_timeout, (tuple, float,)):
+            try:
+                self.cache_fetch_timeout = float(self.cache_fetch_timeout)
+            except (ValueError, TypeError):
```
john-kurkowski: I'm fine with failing fast here. If people bother to override it, they can bother to use the right type. I'll adjust.

john-kurkowski:
Thanks for the contribution and patience!

CI failures are due to Python 3.3 end-of-life and travis-ci/travis-ci#8081. I can pursue a fix to those outside this PR.

TylerLubeck:
Awesome, thank you!!

PR #140, opened 2017-10-11
Closes session in find_first_response

pin3da:
This resolves the warning: `tldextract/tldextract.py:278: ResourceWarning: unclosed`

### DIFF
--- a/tldextract/remote.py
+++ b/tldextract/remote.py
@@ -42,7 +42,9 @@ def find_first_response(urls, cache_fetch_timeout=None):
                 url
             )
         else:
-            return _decode_utf8(text)
+            ans = _decode_utf8(text)
+            session.close()
+            return ans
 
     LOG.error(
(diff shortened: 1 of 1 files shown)


tldextract/remote.py
```diff
@@ -38,7 +38,9 @@ def find_first_response(urls):
         except requests.exceptions.RequestException as ree:
             LOG.error('Exception reading Public Suffix List url ' + url + ' - ' + str(ree) + '.')
         else:
```
john-kurkowski: This doesn't close `session` if it was opened but there's an exception. I'll switch to using it as a context manager. This should still be readable, without crazy `try` nesting.

john-kurkowski:
Rebased Atop Latest Upstream Branch (RALUB).

john-kurkowski:
Thanks for tracking this down! CI failures are also on `master`. They're not to do with this PR. Merging.

PR #150, opened 2018-06-13
Add option to update TLDs list over proxy

czechnology:
I am in a scenario where the Internet can only be accessed over a proxy, yet that proxy is not meant to be in env. vars.
This PR adds option to update the TLDs over a proxy, without unnecessarily cluttering the main class initializer with proxies.

### DIFF
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -3,6 +3,7 @@
 
 import sys
 
+import requests
 import responses
 import tldextract
 from .helpers import temporary_file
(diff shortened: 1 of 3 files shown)


hangtwenty:
First reaction was - hmm I wish we did not have to cascade this argument through a number of places. But I do agree it's ideal not to have it in the constructor. Just wondering if at this point, should it support passing in a `Session` instance instead of this one argument? 

@john-kurkowski thoughts?

john-kurkowski:
I agree @hangtwenty. Let's not maintain our own `requests` shim language, always playing catchup. Let the caller optionally decide how we should make requests, with whatever `requests` options they want.

john-kurkowski:
Closing due to inactivity. Concentrate on #158.

PR #152, opened 2018-06-18
Drop support for EOL Python 2.6

hugovk:
Python 2.6 is EOL and no longer receiving security updates (or any updates) from the core Python team.

It's also little used.

Here's the pip installs for `tldextract` from PyPI for May 2018:

| python_version | percent | download_count |
| -------------- | ------: | -------------: |
| 2.7            |  65.11% |         76,370 |
| 3.6            |  14.82% |         17,386 |
| 3.5            |  14.82% |         17,378 |
| 3.4            |   5.16% |          6,050 |
| 3.7            |   0.07% |             84 |
| 2.6            |   0.02% |             23 |
| 3.3            |   0.00% |              1 |
| Total          |         |        117,292 |

Source: `pypinfo --start-date 2018-05-01 --end-date 2018-05-31 --percent --markdown tldextract pyversion`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
     packages=['tldextract'],
     include_package_data=True,
     long_description=LONG_DESCRIPTION,
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
     classifiers=[
         "Development Status :: 5 - Production/Stable",
         "Topic :: Utilities",
--- a/setup.py
+++ b/setup.py
@@ -26,7 +26,6 @@ You can optionally support the Public Suffix List's private domains as well.
 """
 
 import re
-import sys
 from setuptools import setup
 
 # I don't want to learn reStructuredText right now, so strip Markdown links
@@ -35,8 +34,6 @@ LONG_DESCRIPTION_MD = __doc__
 LONG_DESCRIPTION = re.sub(r'(?s)\[(.*?)\]\((http.*?)\)', r'\1', LONG_DESCRIPTION_MD)
 
 INSTALL_REQUIRES = ["setuptools", "idna", "requests>=2.1.0", "requests-file>=1.4"]
-if (2, 7) > sys.version_info:
-    INSTALL_REQUIRES.append("argparse>=1.2.1")
 
 setup(
     name="tldextract",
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -54,12 +51,12 @@
     packages=['tldextract'],
     include_package_data=True,
     long_description=LONG_DESCRIPTION,
+    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
```
hangtwenty: This is a fairly recent arg, if someone is using Ye Olde 2.6 they'd probably have old enough libraries that it would not stop them

image

https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

setup.py
```diff
@@ -35,8 +34,6 @@
 LONG_DESCRIPTION = re.sub(r'(?s)\[(.*?)\]\((http.*?)\)', r'\1', LONG_DESCRIPTION_MD)
 
 INSTALL_REQUIRES = ["setuptools", "idna", "requests>=2.1.0", "requests-file>=1.4"]
-if (2, 7) > sys.version_info:
-    INSTALL_REQUIRES.append("argparse>=1.2.1")
```
hangtwenty: @hugovk so in light of the other comment, would you mind, instead of deleting line 39 here, actually raise an error like `RuntimeError`? to hard-block people from installing if they are on 2.6.

hugovk: Done! Or I've put the warning right near the top of the file to make sure it's executed right away, and before any potential syntax errors in later code that may only occur on 2.6.

setup.py
hugovk: Fair point.

In the worst case, they have ancient pip and it will ignore this argument.

In the best case, they have modern pip and it will read the metadata and it'll automatically install the last version of this library to support 2.6.

I recommend using latest pip, setuptools (and wheel and twine) when releasing this project to make sure the argument is available to those with modern pip.

setup.py
```diff
@@ -29,14 +29,17 @@
 import sys
 from setuptools import setup
 
+if sys.version_info < (2, 7):
+    raise RuntimeError("Python 2.6 is EOL and no longer supported. "
+                       "Please upgrade your Python or use an older "
+                       "version of tldextract.")
```
hangtwenty: 👍

hangtwenty:
@john-kurkowski [not urgent but nice to have] any reason not to merge?

john-kurkowski:
Merged in ea32c9c04cb4d70ed30b547f3971c85760cad560. Thanks all! 🙌

PR #154, opened 2018-09-12
add a default value to TLDEXTRACT_CACHE_TIMEOUT

guoyuqi:
if no default value defined with TLDEXTRACT_CACHE_TIMEOUT, tldextract maybe hang indefinitely since the requests never go through.

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -92,7 +92,7 @@ LOG = logging.getLogger("tldextract")
 
 CACHE_FILE_DEFAULT = os.path.join(os.path.dirname(__file__), '.tld_set')
 CACHE_FILE = os.path.expanduser(os.environ.get("TLDEXTRACT_CACHE", CACHE_FILE_DEFAULT))
-CACHE_TIMEOUT = os.environ.get('TLDEXTRACT_CACHE_TIMEOUT')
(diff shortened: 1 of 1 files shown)


hangtwenty:
@guoyuqi This is a "hmmmmm" one because ... I agree for my _personal_ preferences that I like when things don't timeout (would rather just get an error, and have my calling code be able to do retries). but kind of also thinking that `requests`, by design, [does not specify a default timeout](http://docs.python-requests.org/en/master/user/advanced/#timeouts). so what's more astonishing to the general user, having it hang, or having it deviate from the default `requests` would use? Not sure!

the compromise solution is seemingly: have a default timeout but have it be very long. like 5 minutes. 

what do you think @guoyuqi ?

since there is already a configuration variable for it though, there is a workaround, so i'm marking it as lower-priority than the other tickets

john-kurkowski:
I'd defer to `requests`. Expect this library to default the same as the popular, familiar Python library.

hangtwenty:
yeah - Principle of Least Astonishment - do what someone would expect for a library that wraps `requests`

hangtwenty:
Would be good to add a note to the README though since it does not point out this environment variable being available. @guoyuqi would you want to open a PR with a readme update adding to the 'caching' section?

hangtwenty:
created #159 about updating README, closing this one

hangtwenty:
thanks again @guoyuqi !

PR #163, opened 2019-07-11
Use pkgutil instead of pkg_resources to read the tld snapshot.

benjaminp:
pkgutil is in the standard library, and using it leads to less code.


john-kurkowski:
Neat. Thanks!

PR #187, opened 2020-02-03
Relocate version number, to avoid costly imports

### LINKED ISSUE #180: Unnecessarily costly import
The main `__init__.py` import the version from the `cli` module. The `cli` module has *expensive* import-time side effects, which leak into any import. The result is immediately visible on the following flame-graph gathered with py-spy:

Screen Shot 2019-12-04 at 3 10 55 PM

This is a whopping 1.8s added to startup time!

This should be easy to fix by moving the version field to a separate file which is imported by both the `cli` and the main API, allowing the main API to stay clear of the `cli` module.

john-kurkowski:
Fixes #180

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
 *.pyc
 *.tld_set
 .cache
+.eggs
 build
 dist
+tldextract/_version.py
 tldextract_app/tldextract
 tldextract_app/web
 tldextract.egg-info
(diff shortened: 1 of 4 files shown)


ddelange:
may i suggest https://github.com/pypa/setuptools_scm/ ?

if you create a tag via local git or via github, your package will be released with proper metadata + importable version specifier. see e.g. setup.py and init py here https://github.com/ddelange/pipgrip
the version will be populated during dist creation and other access to setup(). (e.g pip install -e .

saves a lot of time and one less thing to remember

john-kurkowski:
Thanks for chiming in! I thought there might be a trodden path out there.

PR #197, opened 2020-04-14
 Add an --update_source command line option

chrestomanci:
This change adds an --update_source command line option to the command line interface.

The intended use, is that in production we don't want the delay associated with the default behaviour of updating when the API is called for the first time. This change also makes it possible to change the update URL without changing our main application, just the setup scripts.

This could be a better way to solve the problem that PR #150 is attempting to fix.

The README.md has been updated to document the new feature, but not the changelong.


john-kurkowski:
Holy 2 year bump. 😓  I was previously loathe to expand the CLI, but I now begin to understand the common request. 

For this PR, I fixed conflicts, renamed the flag for consistency, and passed the new list as an argument vs. mutating an object later. I'll noodle on that last one. I don't think the partially constructed object was intentional.

chrestomanci:
Wow! two years!

I don't know if you are waiting for some sort of comment or approval from me, but if you need it, then I would be pleased if you would merge this PR, and while the change in the command line switch will be an inconvenience for me, it is not a reason to delay the merge or edit it further.

Please merge this ASAP.

john-kurkowski:
I'm glad you're still around and find it useful! This PR seems good to me then!

PR #198, opened 2020-04-14
Add code to fold domain names to lower case. Turn it on by default.

chrestomanci:
This change adds an option to fold the output to lower case, and make that the default.

In my view there is no logical reason to preserve the case of input domain names, and it should be considered a bug that the current version of tldextract does that.

Under the current behaviour GOOGLE.COM and google.com are treated as separate domains. This change fixes that.

### DIFF
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -104,13 +104,20 @@ def test_looks_like_ip():
     assert_extract(u'1\xe9', ('', '', u'1\xe9', ''))
 
 
+def test_upper_case_folding():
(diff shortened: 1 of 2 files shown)


john-kurkowski:
Thank you for your interest and effort! 

I haven't heard anyone else mention this problem. It would be a breaking bugfix. There's a trivial, familiar workaround, to pre or post process your strings. I'm wary of another constructor arg (as the broken build reminds us). I don't think I want to follow this route.

Issue #199, opened 2020-04-20
IndexError with malformed punycode text


shermansiu:
In Python 3.6, running
```python3
import tldextract
tldextract.extract("xn--w&")
```

will cause an IndexError.

```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/censored-path/lib/python3.6/site-packages/tldextract/tldextract.py", line 353, in extract
    return TLD_EXTRACTOR(url)
  File "/censored-path/lib/python3.6/site-packages/tldextract/tldextract.py", line 232, in __call__
    translations = [_decode_punycode(label) for label in labels]
  File "/censored-path/lib/python3.6/site-packages/tldextract/tldextract.py", line 232, in <listcomp>
    translations = [_decode_punycode(label) for label in labels]
  File "/censored-path/lib/python3.6/site-packages/tldextract/tldextract.py", line 405, in _decode_punycode
    return idna.decode(label.encode('ascii')).lower()
  File "/censored-path/lib/python3.6/site-packages/idna/core.py", line 392, in decode
    s = ulabel(label)
  File "/censored-path/lib/python3.6/site-packages/idna/core.py", line 310, in ulabel
    label = label.decode('punycode')
IndexError: decoding with 'punycode' codec failed (IndexError: string index out of range)
```

This is due to a bug with Python's punycode decoding, which you can reproduce by running
```python3
"xn--w&".encode("ascii").decode("punycode")
```

This was reported here: https://bugs.python.org/msg295270
and patched here: https://github.com/python/cpython/pull/1986
but the fix was only backported to Python 3.7 and 3.8.

Applying the patch from the Python repository fixes this issue.

Could IndexError also be caught in `tldextract.py:_decode_punycode()` as a workaround?

john-kurkowski:
Thanks for the clear description and fix. Merged.

PR #200, opened 2020-04-21
Catch the IndexError caused by the punycode bug.

shermansiu:
This fix addresses the punycode decoding bug that was referred to in this issue: https://github.com/john-kurkowski/tldextract/issues/199

i.e. Malformed punycode strings like "xn--&" will be treated as if decoding in idna produced an UnicodeError as it should have been.


john-kurkowski:
Reasonable. Thanks!

PR #210, opened 2020-10-21
fix tlds property bug and add missing argument

brycedrennan:
Addresses feedback here:
https://github.com/john-kurkowski/tldextract/pull/207#issuecomment-713748199


john-kurkowski:
Thanks for owning this!

Issue #196, opened 2020-04-11
Deprecate or remove use __file__


jayvdb:
For reading `.tld_set`, `__file__` should be replaced with pkgutil.get_data, which is used for `.tld_set_snapshot`.

The dataset file may have been placed into the package data via another means, and the attribute `__file__` may not be present at all, as it is an optional attribute of the Python module loading system.  tldextract fails badly atm when it is missing.  An example of when it is missing is PyOxidizer, and Nuitka also has similar problems.

When it is missing, or when the package directory is not writable, the cache should be stored in a different location without a warning.

Currently I see the following for every invocation which imports tldextract, because it is installed as a system package used by system apps:

> WARNING unable to cache TLDs in file /usr/lib/python3.8/site-packages/tldextract/.tld_set: [Errno 13] Permission denied: '/usr/lib/python3.8/site-packages/tldextract/.tld_set'

john-kurkowski:
Makes sense. I don't suppose it's as easy as this? Plus changing types from filename strings to read-in-file strings.

```diff
-CACHE_FILE_DEFAULT = os.path.join(os.path.dirname(__file__), '.tld_set')
+CACHE_FILE_DEFAULT = pkgutil.get_data(__name__, '.tld_set')
```

john-kurkowski:
> When it is missing, or when the package directory is not writable, the cache should be stored in a different location without a warning.

Is there a standard, usually writable location for this? Or a library to determine it?

jayvdb:
The xdg standard defines cache directories.

[`appdirs`](https://pypi.org/project/appdirs/) package is typical solution.

john-kurkowski:
That is what I've been looking for. Maybe that's the most robust solution to #9, #149, and this ticket.

merwok:
FYI, `platformdirs` is a replacement for `appdirs` that’s actively maintained and more performant.

PR #211, opened 2020-10-24
Catch permission error when making cache dir, as well as cache file

### LINKED ISSUE #209: 3.0 creates permission error on .suffix_cache
We had a python dependency on a package that had a dependency on tldextract > 2.0.  Our build pipeline has been pulling in tldextract 2.3, but today it pulled in 3.0 and we started getting this exception in our environment:

File "/usr/local/lib/python3.7/site-packages/tldextract/cache.py", line 104, in run_and_cache
cache_filepath = self._key_to_cachefile_path(namespace, key_args)
File "/usr/local/lib/python3.7/site-packages/tldextract/cache.py", line 95, in _key_to_cachefile_path
_make_dir(cache_path)
File "/usr/local/lib/python3.7/site-packages/tldextract/cache.py", line 155, in _make_dir
os.makedirs(os.path.dirname(filename))
File "/usr/local/lib/python3.7/os.py", line 213, in makedirs
makedirs(head, exist_ok=exist_ok)
File "/usr/local/lib/python3.7/os.py", line 223, in makedirs
mkdir(name, mode)
PermissionError: [Errno 13] Permission denied: '/usr/local/lib/python3.7/site-packages/tldextract/.suffix_cache'

john-kurkowski:
Stanches #209.

* Move makedir side-effect from string building helper function into try block
* Duplicate warning from writing a single cache file
* Log this at most once per app
* Remove unused vars


john-kurkowski:
FYI @brycedrennan

brycedrennan:
Ah I hadn't noticed #209 . So it seems some installations have permission to create a file but not a directory. interesting.   I wonder if we should just use a system temp folder instead.  

I'll look more in depth at this tomorrow.

john-kurkowski:
I think it's that creating the directory wasn't wrapped in a `try`. Everything else was. Not about the way user systems distinguish files vs. directories.

System temp folders have been suggested several times in the issue tracker. I was opposed, because I don't want the suffix list to be temporary, evicted at any time by the OS. Maybe I should soften my stance.

My favorite suggestion [is to switch to an XDG standard location](https://github.com/john-kurkowski/tldextract/issues/196#issuecomment-668974052).

john-kurkowski:
Stanches #209 for now. Let's keep that one open for a longer term fix.

brycedrennan:
Sounds good

PR #230, opened 2021-08-12
tests: do not run pylint by default

sbraz:
Hi,
There is already a separate tox environment to run pylint tests. I think it makes sense not to run pylint by default.

While normal tests should be run by downstream (in particular Linux distributions which package tldextract), linter tests should not be run automatically.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -1,2 +1,2 @@
 [pytest]
-addopts = --doctest-modules --pylint
+addopts = --doctest-modules
--- a/tox.ini
+++ b/tox.ini
@@ -10,7 +10,7 @@ deps =
     pytest-pylint
     responses
 
-commands = pytest {posargs}
+commands = pytest --pylint {posargs}
 
 [testenv:codestyle]
 deps = pycodestyle


sbraz:
Hi @john-kurkowski, I am part of the Gentoo Python team and while adding tests for this package, I noticed that they failed without pylint. I understand that pylint is part of your CI and it makes sense that developers run it to check the quality of the code.

However, for downstream (I'm thinking of all distros, such as Fedora, Debian, etc.), tests should only be used to ensure the package works properly. The fact that there are linting issues should not affect the result and we should try to limit the number of dependencies required to run them (and pylint itself has quite a few). This also applies to flake8, mypy, coverage, etc.). 

This is obviously not a deal-breaker as we can always [sed it out in our scripts](https://dev.gentoo.org/~mgorny/python-guide/pytest.html#avoiding-dependencies-on-other-pytest-plugins) but in your case, it didn't seem necessary to run tests for each Python version via tox since you have a dedicated pylint tox environment.

By the way, are you aware of [pytest-enabler](https://pypi.org/project/pytest-enabler/)? It can be used to write configuration files that look like [this](https://github.com/jaraco/jaraco.classes/blob/4fefdd21a4660166cc3136c023e0d3b7604ad573/pyproject.toml).

john-kurkowski:
> However, for downstream (I'm thinking of all distros, such as Fedora, Debian, etc.), tests should only be used to ensure the package works properly. The fact that there are linting issues should not affect the result and we should try to limit the number of dependencies required to run them (and pylint itself has quite a few).

Makes sense to me!

Only one change. In CI, I'd still like to lint across different versions. I recall something about Python version-specific lint errors. I'd like the extra coverage. It'll still spare the downstream distros of the extra dependency.

john-kurkowski:
> In CI, I'd still like to lint across different versions.
> …
> By the way, are you aware of [pytest-enabler](https://pypi.org/project/pytest-enabler/)? It can be used to write configuration files that look like [this](https://github.com/jaraco/jaraco.classes/blob/4fefdd21a4660166cc3136c023e0d3b7604ad573/pyproject.toml).

That seems like an elegant way to keep it in upstream CI and prevent it from reaching downstream! Kinda funny, adding a dependency to avoid dependencies. But I'm willing to give it a try. 🙂

john-kurkowski:
For now, cbd1345 seems simplest, to preserve the CI status quo, yet not force linting in non-Tox, i.e. non-CI environments. What do you think?

sbraz:
Yes that works for me :) You could also create one lint environment per Python version? I don't know how to do that with tox but I guess this is possible.

john-kurkowski:
Yeah, that is nice to have as backup, if my fix turns out to be too simple.

PR #235, opened 2021-10-02
Added documentation for eTLDs

### LINKED ISSUE #234: Better document that this project parses Effective TLDs as the suffix
Per this [question](https://stackoverflow.com/questions/69418561) on Stack Overflow, what tldextract calls the "suffix" is actually an "effective TLD" or "eTLD". In `gov.uk`, `gov` is a second-level domain and `uk` is the `ccTLD`, [per wikipedia](https://en.wikipedia.org/wiki/.uk#Second-level_domains).

Per [Mozilla's documentation on the Public Suffix List](https://wiki.mozilla.org/Public_Suffix_List), an eTLD is
> the highest level at which a domain may be registered for a particular top-level domain

The README should include this term to better describe what it is doing.

pocc:
Fixes #234


Molebogeng1 (approved):

john-kurkowski:
Thank you for this! Went another direction with it. See original issue, #234.

PR #240, opened 2021-12-21
Remove extra backtick in README

AetherUnbound:
This PR removes an extra backtick in the README which was causing some improper rendering.


john-kurkowski:
Thank you!

PR #244, opened 2022-02-04
Add types to the public API

### LINKED ISSUE #189: Type Stubs
It would be awesome if the project could include type stubs for Python 3 to satisfy type check tools such as mypy

hauntsaninja:
Resolves #189

@hangtwenty indicated such a PR would be welcome :-)


john-kurkowski (approved):
LGTM! ❤️ types. Thank you!

PR #245, opened 2022-02-20
Remove py2 tag from wheel

hugovk:
Universal wheels are for supporting Python 2 and 3, and adds the unnecessary `py2` tag to the wheel metadata and filename:

* `tldextract-3.1.2-py2.py3-none-any.whl`

https://pypi.org/project/tldextract/3.1.2/#files

https://wheel.readthedocs.io/en/stable/user_guide.html#building-wheels

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,2 @@
-[bdist_wheel]
-universal = 1
-
 [metadata]
 license_file = LICENSE


john-kurkowski:
Confirmed from [the PyPA's docs on packaging](https://github.com/pypa/packaging.python.org/blob/3bb21ec25325c3b5257eae5469e3e857d041075f/source/guides/distributing-packages-using-setuptools.rst#pure-python-wheels)! Thanks!

PR #247, opened 2022-02-21
Rely on python_requires instead of runtime check

hugovk:
Re: https://github.com/john-kurkowski/tldextract/pull/246#discussion_r810826911

`python_requires` and the removal of the `py2` wheel tag render this unnecessary with modern tools.


john-kurkowski:
🙌

PR #249, opened 2022-02-21
Enforce opinionated code style in CI

john-kurkowski:
I've been manually running Black on files in this project and other projects. I like how it looks, how it removes inconsistencies, and how it removes nitpicks between developers. We've all got bigger fish to fry. This reduces mental burden.

This PR only enforces Black's opinionated code style when running tests. It does not apply it nor enforce it with a e.g. Git pre-commit hook, like using [pre-commit](https://pre-commit.com/).

# Open Issues

* [x] Some way to run automatically apply Black in this project, not just check it
    * It's annoying for CI to bark at you, with no recourse locally. If you don't already have Black wired up in your IDE, independent of this project.
    * [pre-commit](https://pre-commit.com/) is the top candidate. But pre-commit does require a manual install step? Could things be simpler? Just to get this opinionated code style started in this project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,3 +264,12 @@ Run all tests against a specific Python environment configuration:
 tox -l
 tox -e py37
 ```
+
+### Code Style
+
+Automatically format all code:
+
+```zsh
+pip install black
+black .
+```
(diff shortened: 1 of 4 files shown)


john-kurkowski:
I think I addressed the open issue in the PR description by documenting how to run Black directly: install it, then run `black .`.

john-kurkowski:
(Rebased)

PR #255, opened 2022-03-07
Check untyped defs

john-kurkowski:
Catches a few more inconsistencies, and an extraneous code path.

* Fix types
* Fix docs with updated types

### DIFF
--- a/tldextract/cache.py
+++ b/tldextract/cache.py
@@ -118,7 +118,7 @@ class DiskCache:
                         "unable to cache %s.%s in %s. This could refresh the "
                         "Public Suffix List over HTTP every app startup. "
                         "Construct your `TLDExtract` with a writable `cache_dir` or "
-                        "set `cache_dir=False` to silence this warning. %s"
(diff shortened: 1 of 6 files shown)


tldextract/cache.py
```diff
@@ -219,7 +219,8 @@ def _fetch_url(session, url, timeout):
 def _make_cache_key(inputs):
     key = repr(inputs)
     try:
```
john-kurkowski, line 221: @brycedrennan do you remember what this `try` was for? According to the type checker, `repr` above returns a `str`, so `md5(key)` will always raise.

brycedrennan: It's a good question. I'm not sure. the only thing I can think of is Python 2 compatibility.

brycedrennan: yeah I see you removed python 2 compat, so that's likely what this was for

john-kurkowski: Thanks for confirming! Cool. Consider the `try` gone.

PR #256, opened 2022-03-07
Disallow untyped calls and defs

john-kurkowski:
Hot off the heels of #255 …

Continue to allow untyped defs in tests, where the majority of errors are test functions (which pytest picks up by inspecting the function name).

### DIFF
--- a/tldextract/suffix_list.py
+++ b/tldextract/suffix_list.py
@@ -3,9 +3,12 @@
 import logging
 import pkgutil
 import re
+from typing import List, Sequence, Tuple, Union, cast
 
 import requests
 from requests_file import FileAdapter  # type: ignore[import]
+
+from .cache import DiskCache
 
 LOG = logging.getLogger("tldextract")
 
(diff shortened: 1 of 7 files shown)


tldextract/suffix_list.py
```diff
@@ -65,16 +77,20 @@ def get_suffix_lists(cache, urls, cache_fetch_timeout, fallback_to_snapshot):
...
+    fallback_to_snapshot: bool,
+) -> Tuple[List[str], List[str]]:
     """Fetch, parse, and cache the suffix lists"""
 
     try:
         text = find_first_response(cache, urls, cache_fetch_timeout=cache_fetch_timeout)
     except SuffixListNotFound as exc:
         if fallback_to_snapshot:
-            text = pkgutil.get_data("tldextract", ".tld_set_snapshot")
-            if not isinstance(text, str):
```
john-kurkowski, line 76: According to the type of `pkgutil.get_data`, this was always `bytes`, never `str`. 

(Technically `Optional[bytes]`. But we the package maintainers guarantee the inclusion of the file. So force unwrap it, below.)

PR #258, opened 2022-03-17
Fix incorrect namespace used for caching function returns

### LINKED ISSUE #257: "update" caches public suffix list to wrong directory
Hi!

First off, I love `tldextract`, thanks for building it!

The way we use `tldextract` is slightly special, but used to be fully supported by the public API. Our docker containers don't have internet access, so when we build them, we cache the latest public suffix list. When our applications use `tldextract`, we configure it so that it uses the cache, and never needs an internet connection.
Upon upgrading to any 3.* version of `tldextract`, I noticed that the cache was no longer being used to look up information from the public suffix list.

Problem reproduction steps
----------------------------
First, run the command: `tldextract --update --private_domains`
Then create a basic test file:
```python
import os
from tldextract import TLDExtract

extractor = TLDExtract(cache_dir=os.environ["TLDEXTRACT_CACHE"])
extractor("www.google.com")
```

Now, create a conditional breakpoint [here](https://github.com/john-kurkowski/tldextract/blob/master/tldextract/cache.py#L93), where the condition is that `namespace` equals `publicsuffix.org-tlds`.

### Expected behaviour

When running the above program, the break point should be hit, but should not throw a `KeyError`.

### Actual behaviour

The breakpoint is hit once during the `__call__(…)`, and immediately throws a `KeyError` because it can't find the cache file.

Explanation
------------

The method `run_and_cache` accepts a namespace, which is used to calculate the cache file path. But when the file is downloaded, it uses the hardcoded namespace "urls", which places the file in the wrong location.

I'll write a PR that fixes this problem.

nicholas-plutoflume:
This PR fixes #257.

Currently, the namespace is not passed to the function that sets the cache, and instead the literal "urls" is passed. As a result, non-url related functions, such as the function to cache the public suffixes were not cached.

This PR fixes this by using the `namespace` argument which is passed in to the `run_and_cache` function, and adds a regression test.

### DIFF
--- a/tests/test_cache.py
+++ b/tests/test_cache.py
@@ -3,6 +3,7 @@ import os.path
 import sys
 import types
 from typing import Any, cast
+from unittest.mock import Mock
 
 import pytest
 import tldextract.cache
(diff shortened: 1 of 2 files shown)


tests/test_cache.py
```diff
@@ -72,3 +73,29 @@ def test_get_cache_dir(monkeypatch):
...
+    assert call2 == return_value1
+    assert some_fn.call_count == 1
+
+    kwargs2 = {"value": 2}
+    return_value2 = "another return value"
+    some_fn.return_value = return_value2
+
+    call3 = cache.run_and_cache(some_fn, "test_namespace", kwargs2, kwargs2.keys())
+    assert call3 == return_value2
+    assert some_fn.call_count == 2
```
john-kurkowski, line 101: Thanks for the thorough PR! 🙌 

My only nitpick is, I had trouble grokking the test case. We're asserting side effects, so it's going to be an indirect test case no matter what. I expanded coverage of `run_and_cache`'s expected behavior with a `Mock` and asserting `call_count`.  I'm not sure which version I prefer. What do you think?

nicholas-plutoflume: I agree, I like your version more. 👍 
The only trade-off is that we're now testing for implementation details instead of side-effects, but it does make the test easier to understand.

nicholas-plutoflume: (Then again, by passing a function that can fail and asserting that it doesn't throw an exception I guess I was doing both, testing the implementation _and_ side-effects)

john-kurkowski:
(Rebased)

john-kurkowski:
Thanks for the thorough PR and discussion!

PR #260, opened 2022-03-17
Add CLI option to not fall back to the snapshot

### LINKED ISSUE #259: It should be possible to prevent the use of the fallback snapshot when using the CLI
When you initialise the cache, either through the CLI or through `update(fetch_now=True)`, it calls it on a `TLDExtract` object that has the default suffix list and `fallback_to_snapshot` set to `True`. This causes problems if you later try to use a `TLDExtract` object that points to the same cache directory, but with `fallback_to_snapshot` set to `False`.
Because `fallback_to_snapshot` is one of the hashed arguments, it will generate a totally different cache path, and miss the cache set previously through `update`.

I believe it would make sense to make one of two modifications: 
1. Allow the user to explicitly pass through `fallback_to_snapshot` to the `update` function / CLI.
2. Or, ignore the `fallback_to_snapshot` parameter when constructing the cache path.

~I believe option 2 is better for the following reasons:~
- ~It doesn't make sense to call `update` in a scenario where the intent is to use `fallback_to_snapshot`, because you can use `tldextract` without pre-seeding the cache, and just use the snapshot instead.~
- ~The pivotal element should only be the set of public suffix urls. If those change, it makes sense to use a different cache file for lookups.~

~To summarise (assuming the prefix list urls are the same):~
- ~If `fallback_to_snapshot` is turned on or off after you cache items it won't matter because you've already cached the suffix lists, and that's where it should get the data from.~
- ~I can't think of a scenario where you would have the cache off and have two different `TLDExtract` objects (with the same prefix list urls) where for one the fallback is on, for the other it is off.~

~I'll make a PR to fix this using method 2. But I'm curious to hear if you agree with my analysis.~
~

nicholas-plutoflume:
This fixes #259.

It adds a new commandline option which will prevent the lookup / update from using the package's cache.

### DIFF
--- a/tldextract/cli.py
+++ b/tldextract/cli.py
@@ -41,11 +41,19 @@ def main():
         action="store_true",
         help="Include private domains",
     )
+    parser.add_argument(
+        "--no_fallback_to_snapshot",
+        default=True,
+        action="store_false",
+        dest="fallback_to_snapshot",
+        help="Don't fall back to the package's snapshot of the prefix list",
+    )
 
     args = parser.parse_args()
 
     obj_kwargs = {
         "include_psl_private_domains": args.private_domains,
+        "fallback_to_snapshot": args.fallback_to_snapshot,
     }
     if args.cache_dir:
         obj_kwargs["cache_dir"] = args.cache_dir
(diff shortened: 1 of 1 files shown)


john-kurkowski:
Thank you!

Issue #265, opened 2022-06-01
`suffix_list_urls` should be `Optional`


MrRawbin:
According to the README, the `suffix_list_urls` can be set to `None` to prevent live HTTP fetching:

```python3
# extract callable that falls back to the included TLD snapshot, no live HTTP fetching
no_fetch_extract = tldextract.TLDExtract(suffix_list_urls=None)
```
The `TLDExtract` constructor then handles `None` by setting the value to an empty tuple like this:

```python3
suffix_list_urls = suffix_list_urls or ()
```

However, my IDE complains about me using `None` as value for `suffix_list_urls` since it is not `Optional`. A workaround is using an empty tuple as a value myself, but I don't think I should have to.

john-kurkowski:
Thanks for raising this issue! That's a documentation bug. Fixed in cd07547. I prefer the simpler type, as is.

PR #266, opened 2022-06-01
Made suffix_list_urls Optional by type hint

MrRawbin:
Fixes https://github.com/john-kurkowski/tldextract/issues/265


john-kurkowski:
Went another way with this. See discussion on the original issue.

PR #273, opened 2022-09-20
Extract port

### LINKED ISSUE #272: Get the port from the provided URL to extract function
Hi!

I was wondering if it is possible to get the port from the URLs when `extract` function is invoked (or other function). I guess it is not, since I didn't see it in the documentation and I've dug a little bit in the code and I didn't see anything related. I'm using this library in order to obtain URLs from a large list, and use those URLs in order to crawl, so I need the port in case it is defined. In case it is not possible to obtain the port, it is intended to implement this functionality?

```python
>>> tldextract.extract('http://127.0.0.1:8080/deployed/')
ExtractResult(subdomain='', domain='127.0.0.1', suffix='', port='8080')
```

Thank you!

john-kurkowski:
* Add optional `port` as the 4th member of the `ExtractResult` namedtuple
* Deemphasize documentation about rejoining the `ExtractResult` namedtuple members by dot
  * Since port in a URL-like string is optional and joined by colon, not dot

Fixes #272.

# Compatibility

* This breaks the previously documented `".".join(extract_result)` pattern
  * If callers aren't careful, the pattern will produce awkward strings like `"foo.bar.com.None"` or `"foo.bar.com.8080"`, or raise a `TypeError` from trying to join `str` with `Optional[int]`
  * Arguably, relying on an `ExtractResult` being a namedtuple is historical. Still, this PR would complicate things for users relying on that join.

# Other Options

1. Make port an attribute of `ExtractResult` but not a member of the tuple, like `urllib.parse.urlsplit` does
1. Don't merge this; callers can get the port using the standard library (and the port is _validated_)
    * ```py
        >>> import urllib.parse
        >>> print(urllib.parse.urlsplit("//foo.bar.com:8080").port)
        8080
        ```
    * As a consolation to #272, speed up this package when passed a `urllib.parse.{ParseResult,SplitResult}`: see #274

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,20 +56,7 @@ ExtractResult(subdomain='', domain='google', suffix='com')
 ExtractResult(subdomain='google', domain='notavalidsuffix', suffix='')
 
 >>> tldextract.extract('http://127.0.0.1:8080/deployed/')
-ExtractResult(subdomain='', domain='127.0.0.1', suffix='')
-```
-
-If you want to rejoin the whole namedtuple, regardless of whether a subdomain
-or suffix were found:
-
-```python
->>> ext = tldextract.extract('http://127.0.0.1:8080/deployed/')
->>> # this has unwanted dots
->>> '.'.join(ext)
-'.127.0.0.1.'
->>> # join each part only if it's truthy
->>> '.'.join(part for part in ext if part)
-'127.0.0.1'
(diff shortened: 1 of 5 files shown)


john-kurkowski:
After thinking about it, this library is focused on domain names, not every component of a URL. I defer URL parsing to Python's standard lib.

PR #279, opened 2023-01-26
Fix pyright not finding tldextract public interface

agateau-gg:
When I run [pyright](https://github.com/microsoft/pyright) against this code (inspired from the README):

```python
import tldextract

print(tldextract.extract('http://forums.news.cnn.com/'))

```

it complains like this:

```
pyright 1.1.291
/home/agateau/tmp/tst.py
  /home/agateau/tmp/tst.py:3:18 - error: "extract" is not exported from module "tldextract" (reportPrivateImportUsage)
```

Listing all symbols in `__all__` fixes the problem.

### DIFF
--- a/tldextract/__init__.py
+++ b/tldextract/__init__.py
@@ -4,3 +4,9 @@ from . import _version
 from .tldextract import TLDExtract, extract
 
 __version__: str = _version.version
+
+__all__ = [
+    "extract",
+    "TLDExtract",
+    "__version__",
+]


john-kurkowski:
Nice! I was in this file recently trying out [Ruff](https://github.com/charliermarsh/ruff). It was complaining about the file's unused imports. Anyway, explicit is better than implicit. Thank you!

PR #280, opened 2023-03-29
Use SPDX license identifier

vargenau:
Use SPDX license identifier: BSD-3-Clause
This will help tools to produce valid SPDX.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         "exceptions. You can optionally support the Public Suffix "
         "List's private domains as well."
     ),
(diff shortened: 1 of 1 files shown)


john-kurkowski:
Using a standard sounds good. Thank you! 

(I wish Python's [own documentation on `license`](https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#license) called out SPDX as possible values. Instead, they say license specifics should be set via the `classifiers` field, whose possible values are not exhaustive and are Python-specific.)

PR #286, opened 2023-05-18
Upgrade to newer subscriptable types

john-kurkowski:
To keep up with the Python language Joneses.

# Changes

* Use minimum Python for static checks
* Support future syntax

### DIFF
--- a/tldextract/cache.py
+++ b/tldextract/cache.py
@@ -1,4 +1,6 @@
 """Helpers."""
+from __future__ import annotations
+
 import errno
 import hashlib
 import json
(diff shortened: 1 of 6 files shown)


tldextract/cache.py
```diff
@@ -113,7 +110,7 @@ def get(self, namespace: str, key: Union[str, Dict[str, Hashable]]) -> object:
             raise KeyError("namespace: " + namespace + " key: " + repr(key)) from None
 
     def set(
-        self, namespace: str, key: Union[str, Dict[str, Hashable]], value: object
+        self, namespace: str, key: str | dict[str, Hashable], value: object
```
elliotwutingfeng, line 113: This syntax `|` requires [Python 3.10 or newer](https://peps.python.org/pep-0604)

tox.ini
```diff
@@ -29,6 +31,7 @@ deps =
 commands = pytest --pylint -m pylint {posargs}
 
 [testenv:typecheck]
+basepython = python3.7
```
john-kurkowski, line 34: I would've thought the newer syntax in this PR would break on older Python versions, and this project wouldn't be able to adopt the syntax yet. However, the `from __future__ import annotations` at the top of source files that use the syntax seems to allow the syntax, as far back as Python 3.7.

PR #292, opened 2023-05-25
Accept only 4 decimal octet IPv4 addresses. Support IPv4 addresses with unicode dots.

### LINKED ISSUE #287: IPv4 with non-ASCII dots
Should IPv4 addresses with non-ASCII dots be accepted?

Example: `http://127\u30020\uff610\u002e1/foo/bar`

### LINKED ISSUE #290: 1,2,3-octet/hexadecimal hostnames detected as IPv4 addresses
The following inputs are recognized as IPv4 addresses due to the use of **socket.inet_aton()**.

`1.1.1` -> domain parsed as `1.1.1`
`1.1` -> domain parsed as `1.1`
`1` -> domain parsed as `1` (output is still correct nonetheless)

The above is legacy behavior from [UNIX's inet_aton](https://manpages.debian.org/bullseye/manpages-dev/inet.3.en.html) for [classful networks](https://en.wikipedia.org/wiki/Classful_network), a network addressing architecture made obsolete in 1993.

`01.01.01.01` -> domain parsed as `01.01.01.01`
`01.01.01` -> domain parsed as `01.01.01`
`01.01` -> domain parsed as `01.01`
`01` -> domain parsed as `01` (output is still correct nonetheless)

`0x1.0x1.0x1.0x1` -> domain parsed as `0x1.0x1.0x1.0x1`
`0x1.0x1.0x1` -> domain parsed as `0x1.0x1.0x1`
`0x1.0x1` -> domain parsed as `0x1.0x1`
`0x1` -> domain parsed as `0x1` (output is still correct nonetheless)

Given that tldextract's regex-based **ipv4()** function only recognizes IPv4 addresses with 4 decimal octets without zero padding, this is probably a bug.

It can be fixed by using **socket.inet_pton()** in **looks_like_ip()** instead of **socket.inet_aton()**. However, it is only supported on Unix/Unix-Like/Windows systems. Some of [these](https://www.python.org/download/other) systems do not.

A more portable fix would be using **ipaddress.IPv4Address**, though it is much slower.

If **suffix_index == len(labels) == 4**, are there any edge cases not covered by **IP_RE**?

elliotwutingfeng:
## Changes

- IPv4 addresses with unicode dots are now recognized. Closes #287
- IPv4 addresses must have 4 decimal octets. Closes #290

**ipaddress.IPv4Address** was not used as fallback because it is 5 times slower than regex, and its behavior is [inconsistent](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address) across Python versions 3.8 and 3.9 (w.r.t. recognition of leading zeroes). For example, on Raspberry Pi OS Bullseye, the latest Python version is 3.9.2-3, which allows leading zeroes. To update to the latest patch version of Python 3.9, one has to manually install from source.

### DIFF
--- a/tldextract/remote.py
+++ b/tldextract/remote.py
@@ -1,11 +1,16 @@
 """tldextract helpers for testing and fetching remote resources."""
 
-try:
-    from socket import inet_pton, AF_INET  # Availability: Unix, Windows.
-except ImportError:
-    inet_pton = None  # type: ignore
+from __future__ import annotations
+
 import re
+from collections.abc import Callable
 from urllib.parse import scheme_chars
+
+inet_pton: Callable[[int, str], bytes] | None
+try:
+    from socket import AF_INET, inet_pton  # Availability: Unix, Windows.
+except ImportError:
+    inet_pton = None
 
 IP_RE = re.compile(
     r"^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.)"
--- a/tldextract/remote.py
+++ b/tldextract/remote.py
@@ -1,7 +1,10 @@
 """tldextract helpers for testing and fetching remote resources."""
 
+try:
+    from socket import inet_pton, AF_INET  # Availability: Unix, Windows.
+except ImportError:
+    inet_pton = None  # type: ignore
 import re
-import socket
 from urllib.parse import scheme_chars
 
 IP_RE = re.compile(
@@ -49,13 +52,11 @@ def looks_like_ip(maybe_ip: str) -> bool:
     if not maybe_ip[0].isdigit():
         return False
 
-    try:
-        socket.inet_aton(maybe_ip)
-        return True
-    except (AttributeError, UnicodeError):
-        if IP_RE.match(maybe_ip):
+    if inet_pton is not None:
+        try:
+            inet_pton(AF_INET, maybe_ip)
             return True
-    except OSError:
-        pass
+        except OSError:
+            return False
 
-    return False
+    return IP_RE.match(maybe_ip) is not None
--- a/tldextract/remote.py
+++ b/tldextract/remote.py
@@ -52,16 +52,17 @@ def _schemeless_url(url: str) -> str:
     return url[double_slashes_start + 2 :]
 
 
-def looks_like_ip(maybe_ip: str) -> bool:
+def looks_like_ip(
+    maybe_ip: str, pton: Callable[[int, str], bytes] | None = inet_pton
+) -> bool:
     """Check whether the given str looks like an IP address."""
     if not maybe_ip[0].isdigit():
         return False
 
-    if inet_pton is not None:
+    if pton is not None:
         try:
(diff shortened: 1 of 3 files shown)


tldextract/remote.py
```diff
@@ -49,13 +57,11 @@ def looks_like_ip(maybe_ip: str) -> bool:
     if not maybe_ip[0].isdigit():
         return False
 
-    try:
-        socket.inet_aton(maybe_ip)
-        return True
-    except (AttributeError, UnicodeError):
-        if IP_RE.match(maybe_ip):
+    if inet_pton is not None:
+        try:
```
john-kurkowski, line 61: If I comment out this block, i.e. `inet_pton` is always unavailable, the test suite still passes. Can you add an assert demonstrating the use of `inet_pton`?

```diff
diff --git a/tldextract/remote.py b/tldextract/remote.py
index 1c6cad2..60ca31a 100644
--- a/tldextract/remote.py
+++ b/tldextract/remote.py
@@ -3,14 +3,15 @@
 from __future__ import annotations
 
 import re
-from collections.abc import Callable
+
+# from collections.abc import Callable
 from urllib.parse import scheme_chars
 
-inet_pton: Callable[[int, str], bytes] | None
-try:
-    from socket import AF_INET, inet_pton  # Availability: Unix, Windows.
-except ImportError:
-    inet_pton = None
+# inet_pton: Callable[[int, str], bytes] | None
+# try:
+#     from socket import AF_INET, inet_pton  # Availability: Unix, Windows.
+# except ImportError:
+#     inet_pton = None
 
 IP_RE = re.compile(
     r"^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.)"
@@ -57,11 +58,11 @@ def looks_like_ip(maybe_ip: str) -> bool:
     if not maybe_ip[0].isdigit():
         return False
 
-    if inet_pton is not None:
-        try:
-            inet_pton(AF_INET, maybe_ip)
-            return True
-        except OSError:
-            return False
+    # if inet_pton is not None:
+    #     try:
+    #         inet_pton(AF_INET, maybe_ip)
+    #         return True
+    #     except OSError:
+    #         return False
 
     return IP_RE.match(maybe_ip) is not None
```

john-kurkowski (commented):
Good stuff.

tldextract/remote.py
elliotwutingfeng: Do you means something like `assert inet_pton is None`?

john-kurkowski: Oh, that's an improvement!

What I originally meant was testing `inet_aton` vs `inet_pton`. But that's already covered by an assert you added later in this file. With `inet_aton`, the assert fails, with `inet_pton` it passes.

```
    assert_extract(
        "http://127.0.0/foo/bar", ("", "127.0", "0", ""), expected_ip_data=""
    )
```

I checked out master with your new tests and figured that out. So, this PR all good!

john-kurkowski:
Thank you!

PR #293, opened 2023-06-07
Reject IPv4 addresses with trailing whitespaces + non-whitespaces

elliotwutingfeng:
### Proposed changes

- Reject edge cases where inputs like `1.1.1.1 com` or `1.1.1.1\rsomething` are parsed as IPv4 addresses. **fullmatch** is about 10% slower than **match** but it'll only affect systems that do not support **pton** (non-Unix/Windows).
- Added some tests for test coverage.
- Fixed a minor typo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ tldextract.egg-info
 tldextract/.suffix_cache/*
 .tox
 .pytest_cache
+.coverage
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,3 @@ tldextract.egg-info
 tldextract/.suffix_cache/*
 .tox
 .pytest_cache
-.coverage
--- a/tests/cli_test.py
+++ b/tests/cli_test.py
@@ -38,8 +38,16 @@ def test_cli_posargs(capsys, monkeypatch):
 
 def test_cli_namedargs(capsys, monkeypatch):
     monkeypatch.setattr(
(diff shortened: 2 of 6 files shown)


tests/cli_test.py
```diff
@@ -33,3 +34,24 @@ def test_cli_posargs(capsys, monkeypatch):
     stdout, stderr = capsys.readouterr()
     assert not stderr
     assert stdout == " example com\n bbc co.uk\nforums bbc co.uk\n"
+
+
+def test_cli_namedargs(capsys, monkeypatch):
```
john-kurkowski, line 39: I appreciate the increase in test coverage. I'm going to cherry-pick it outside of this PR, because it distracts from the main theme of the PR, the IP bugfix.

.gitignore
```diff
@@ -9,3 +9,4 @@ tldextract.egg-info
 tldextract/.suffix_cache/*
 .tox
 .pytest_cache
+.coverage
```
john-kurkowski, line 12: This project doesn't use any coverage tools, so this line is a little confusing. The line might belong in your personal .gitignore? I'm going to omit it from the PR. Although I would love coverage +/- tracking on PRs someday… 😄

john-kurkowski (commented):
Thank you!

.gitignore
elliotwutingfeng: Agreed, it would be better in another future PR.

PR #298, opened 2023-07-07
Add support for IPv6 addresses

### LINKED ISSUE #263: IPv6 addresses are not handled
For URLs using IPv4 addresses, the host address gets extracted correctly using `.domain` but `.fqdn` gives the empty string:
```python
>>> tldextract.extract("https://127.0.0.1:1234/foobar").domain
'127.0.0.1'
>>> tldextract.extract("https://127.0.0.1:1234/foobar").fqdn
''
```
For URLs using IPv6 addresses, neither method extracts the host address correctly:
```python
>>> tldextract.extract("https://[::1]:1234/foobar").domain
'['
>>> tldextract.extract("https://[::1]:1234/foobar").fqdn
'' 
>>> tldextract.extract("https://[FEC0:0000:0000:0000:0000:0000:0000:0001]:1234/foobar").domain
'[FEC0'
>>> tldextract.extract("https://[FEC0:0000:0000:0000:0000:0000:0000:0001]:1234/foobar").fqdn
''
```

This was tested using `tldextract` version 3.2.1 on Python 3.9.12

elliotwutingfeng:
Closes #263

- This could possibly break an unusual use case where square brackets are considered valid domain name labels, such as `sub[.example.com` -> `sub[` + `example` + `com`. It's not compliant with [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986#section-3) to begin with so it probably isn't an issue.

- If the IPv6 literal contains a trailing IPv4 address, non-ASCII dots are allowed, similar to behaviour in the Chrome and Edge web browsers. Firefox however does not allow non-ASCII dots in IPv6 literals.

- I don't see any performance regressions as of yet.

- Pure IPv6 addresses without square brackets are deemed invalid because they are invalid in the context of URLs.

### DIFF
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -300,11 +300,15 @@ def test_username():
         expected_ip_data="127.0.0.1",
     )
     assert_extract(
+        "https://apple:pass@[::]:50/a",
+        ("", "", "[::]", ""),
+        expected_ipv6_data="::",
+    )
+    assert_extract(
         "https://apple:pass@[aBcD:ef01:2345:6789:aBcD:ef01:127.0.0.1]:50/a",
         ("", "", "[aBcD:ef01:2345:6789:aBcD:ef01:127.0.0.1]", ""),
         expected_ipv6_data="aBcD:ef01:2345:6789:aBcD:ef01:127.0.0.1",
     )
-
 
 def test_query_fragment():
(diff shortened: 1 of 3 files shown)


john-kurkowski:
Amazing! I'll take a look soon.

john-kurkowski:
✨ Beautiful. Well integrated with what's already there.

Issue #138, opened 2017-09-19
Semantics of `registered_domain` property for private domains


tuler:
Suppose the following url: `tuler.github.io`
`github.io` is a private domain in the PSL.

When parsed with `include_psl_private_domains=True` we get `subdomain=''`, `domain=tuler`, `suffix=github.io`.

The `registered_domain` property just joins `domain` and `suffix`, giving me `tuler.github.io`, but IMHO it still should be `github.io`, as this is the domain registered with the registrar, and can be found in a whois query.

One problem to implement this is that when a URL is parsed, we can't know if the parsed domain is a private domain or a ICANN domain, because this is not kept internally when the PSL is read.

Any thoughts?

john-kurkowski:
(Note to self, if we need to track public vs. private at runtime, #66 is a requirement.)

john-kurkowski:
Yeah, I bet most will associate it with registrar registration, as you have.

In my mind, tldextract has been consistent, working as designed, via a more abstract interpretation of "registered." Excluding private domains, GitHub registered [github.io](//github.io) with a registrar, who controlled the domain. Including private domains, GitHub user tuler "registered" [tuler.github.io](//tuler.github.io) with GitHub, who controlled the domain.

I have no strong evidence if my interpretation is broadly useful. It was for a very specific case, when I originally wrote this lib. Or maybe both interpretations are useful.

tuler:
I see your point.

Nonetheless, keeping runtime information regarding each domain from the PSL can be useful to handle this appropriately by the application. Something like a `is_private` method, or a `is_private` flag added to the `ExtractResult`.

john-kurkowski:
Yes, at the very least we should do #66 and expose `is_private`.

I'd then consider a new `registered_domain`-like that was constant in the face of not/private. Just needs a new name.

Renaming today's `registered_domain` is also a possibility, but then we're burdened with backwards compat and legacy association with today's wording.

john-kurkowski:
The PR for #66 currently tracks the source of an extraction, whether the official public suffix list, the private domains in the public suffix list, or user-provided extra suffixes. We haven't figured out how to expose that yet. It's tricky, since it's a `namedtuple`.

gregtws:
Perhaps adding new APIs using terminology reflected by the PSL and libraries like Google Guavas' [InternetDomainName](https://guava.dev/releases/snapshot-jre/api/docs/com/google/common/net/InternetDomainName.html) would help?

[...]

PR #300, opened 2023-07-12
Accept next largest suffix if entire URL is private suffix

### LINKED ISSUE #178: Incorrect empty FQDN when no subdomain and domain
When the extract result's subdomain and domain are both empty, it will return an incorrect empty FQDN.

```python
Python 3.6.7 (default, Oct 22 2018, 11:32:17) 
[GCC 8.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tldextract
>>> tldextract.__version__
'2.2.0'
>>> tldextract.extract('http://s3.amazonaws.com/')
ExtractResult(subdomain='', domain='', suffix='s3.amazonaws.com')
>>> tldextract.extract('http://s3.amazonaws.com/').fqdn
''
>>> tldextract.extract('http://blogspot.com/')
ExtractResult(subdomain='', domain='', suffix='blogspot.com')
>>> tldextract.extract('http://blogspot.com/').fqdn
''
```

#174 would be a related issue but it's not the same.

Is this check really necessary?
https://github.com/john-kurkowski/tldextract/blob/d6574145f76d916ce978eeb898c9e022cf31b87f/tldextract/tldextract.py#L121-L124

elliotwutingfeng:
Closes #178

### Changes

- If private suffixes are enabled, when the entire URL is a private suffix, the parser will accept the next largest suffix (can be either public or private). This is possible since every private suffix is guaranteed to contain at least one public suffix.
- This is a breaking change so perhaps we should use a flag to optionally enable this behaviour.

### Examples

- Entire URL is private suffix **icann.compute.amazonaws.com**
- Next largest suffix is **com**

`icann.compute.amazonaws.com` | subdomain="icann.compute", domain="amazonaws", suffix="com"
`another.icann.compute.amazonaws.com` | subdomain="", domain="another", suffix="icann.compute.amazonaws.com"

- Entire URL is private suffix **s3.dualstack.us-east-1.amazonaws.com**.
- Next largest suffix is **us-east-1.amazonaws.com** (which happens to be a private suffix too)

`s3.dualstack.us-east-1.amazonaws.com` | subdomain="s3", domain="dualstack", suffix="us-east-1.amazonaws.com"
`another.s3.dualstack.us-east-1.amazonaws.com` | subdomain="", domain="another", suffix=s3.dualstack.us-east-1.amazonaws.com"

### Misc

This can be a step towards fixing #138. Would need a consensus on whether **registered_domain** should strictly refer to domain + public suffix regardless of whether private suffixes are enabled.

### DIFF
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -475,6 +475,11 @@ def test_global_extract():
     ) == ExtractResult(
         subdomain="", domain="another", suffix="icann.compute.amazonaws.com"
     )
+    assert tldextract.extract(
+        "another.s3.dualstack.us-east-1.amazonaws.com", include_psl_private_domains=True
+    ) == ExtractResult(
+        subdomain="", domain="another", suffix="s3.dualstack.us-east-1.amazonaws.com"
+    )
 
     assert tldextract.extract(
         "s3.ap-south-1.amazonaws.com", include_psl_private_domains=True
--- a/tests/main_test.py
+++ b/tests/main_test.py
@@ -448,9 +448,6 @@ def test_global_extract():
         "foo.blogspot.com", include_psl_private_domains=True
     ) == ExtractResult(subdomain="", domain="foo", suffix="blogspot.com")
     assert tldextract.extract(
-        "s3.ap-south-1.amazonaws.com", include_psl_private_domains=True
-    ) == ExtractResult(subdomain="", domain="", suffix="s3.ap-south-1.amazonaws.com")
-    assert tldextract.extract(
         "the-quick-brown-fox.ap-south-1.amazonaws.com", include_psl_private_domains=True
     ) == ExtractResult(
         subdomain="the-quick-brown-fox.ap-south-1", domain="amazonaws", suffix="com"
@@ -461,9 +458,6 @@ def test_global_extract():
     assert tldextract.extract(
         "amazonaws.com", include_psl_private_domains=True
     ) == ExtractResult(subdomain="", domain="amazonaws", suffix="com")
-    assert tldextract.extract(
-        "s3.cn-north-1.amazonaws.com.cn", include_psl_private_domains=True
-    ) == ExtractResult(subdomain="", domain="", suffix="s3.cn-north-1.amazonaws.com.cn")
     assert tldextract.extract(
         "the-quick-brown-fox.cn-north-1.amazonaws.com.cn",
         include_psl_private_domains=True,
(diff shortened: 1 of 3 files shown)


john-kurkowski:
Is the point of this PR mainly to fix the `.fqdn` property in #178? Is there a way to fix the property without a breaking change? I don't like that extracting a public suffix now gives a different result than extracting a private suffix. I like treating public and private (when enabled) as similarly as possible.

```python
# entire URL is a public suffix
>>> tldextract.extract("gov.au", include_psl_private_domains=True)
ExtractResult(subdomain='', domain='', suffix='gov.au') # not domain='gov', suffix='au'?

# entire URL is a private suffix
>>> tldextract.extract("blogspot.com", include_psl_private_domains=True)
ExtractResult(subdomain='', domain='blogspot', suffix='com')
```

Also, I don't know how useful it is to check if an entire URL was a suffix, but you could check that before this PR. After this PR, I don't think you can.

```py
>>> result = tldextract.extract("blogspot.com", include_psl_private_domains=True)
>>> is_suffix = result.suffix and not (result.subdomain or result.domain)
```

elliotwutingfeng:
> Is there a way to fix the property without a breaking change?

This can be done with an opt-in flag (not yet implemented as of now) to enable the new behaviour.

`tldextract.extract("blogspot.com", include_psl_private_domains=False, opt_in_flag_name=False)`

If the caller doesn't set this flag to True, the old behaviour applies. If you find this approach feasible, what would be an appropriate name to replace `opt_in_flag_name`?

john-kurkowski:
What desirable new behavior would the flag let users opt into? Is the flag just a bugfix for #178? If so, breaking changes, new configuration options, and something the caller has to do all seem like a chainsaw, for what calls for a knife. I'm afraid to maintain 2 different string splitting experiences or more options to the already too many to construct a `TLDExtract`.

elliotwutingfeng:
Yes, this flag is just a bugfix for #178. If an opt-in flag is used, none of the existing code out there should be affected, and hence it would not be a breaking change.

However I do agree that this complicates the string splitting algorithm; having to handle 2 different cases. If it is too much, I think #178 (which is currently still open) should be closed as "not planned".

john-kurkowski:
> After re-reading the issue, I realised your solution of using the is_private property is simpler and there won't be a need ti opt-in, but it would change the output of FQDN for the private domains.

You got it. That approach is what I'm interested in. #178 is a legit bug, so I'll keep it open for now. The FQDN of private domains was never intended to be blank, so changing that output is desirable.

john-kurkowski:
This PR changed a lot since my last comment! I think it's much closer to the solution I was seeking in my last comment. Nice! 

My only remaining concern is the semi-breaking change of adding another field to this project's `class ExtractResult(NamedTuple)`. This project's docs previously advocated iterating through the fields. With this PR, anybody still iterating through the namedtuple is going to build some weird strings or more likely raise a `TypeError`. That is, without guards e.g. `isinstance(str)` in this PR's updated docs. I noted a similar compatibility issue on the abandoned #273.

I wonder if there's a better way? Do we add an attribute to the namedtuple, instead of extending its fields? Do we bump the breaking version number of the project? Do we move away from namedtuple?

I don't want to make this PR's fix go on much longer. Just noodling what compatibility I'm comfortable with. 🤔

john-kurkowski:
> Do we add an attribute to the namedtuple, instead of extending its fields?

I've played around with this, and it doesn't feel viable. `__slots__` and `__new__` cannot be overridden on a `namedtuple`.

> Make … an attribute of `ExtractResult` but not a member of the tuple, like `urllib.parse.urlsplit` does

I still haven't investigated what the stdlib is doing differently in e.g. `urllib.parse.urlsplit`'s example. I'm not sure it's worth the hoop jumping for this project.

> Do we bump the breaking version number of the project?

Doing this for every field added to the namedtuple seems like a lot. I'd just steer people away from iterating over all fields, unless they really know what they're doing. 

All this to say, I'm leaning toward merging this as is, releasing, and seeing how niche the compatibility issues are.

banagale:
> My only remaining concern is the semi-breaking change of adding another field to this project's `class ExtractResult(NamedTuple)`. This project's docs previously advocated iterating through the fields. With this PR, anybody still iterating through the namedtuple is going to build some weird strings or more likely raise a `TypeError`. That is, without guards e.g. `isinstance(str)` in this PR's updated docs. I noted a similar compatibility issue on the abandoned #273.

I had some code using `tldextract.extract(url)[1:]` and hit this TypeError! Came across it despite not having test coverage on this portion of the product. 

Glad this package is getting improved and am okay with the breaking change, 

I don't know if I would have caught it, but if this had kicked the project up to 4.0.0 (to help highlight the breaking change). I _might_ have checked the release notes more carefully. I realize not everyone uses that convention, and that this is not such a deeply breaking change to possibly warrant it. But throwing it out there nonetheless.

As it was, I just updated to `site_str: str = '.'.join(tldextract.extract(url)[1:3])` and got back to it. 

Thanks for any and all efforts to maintain and further extend this great package.

john-kurkowski:
@banagale thank you for weighing in! Per #305, 3.6.0 is yanked and republished as 4.0.0. Your tuple slicing code will work in 3.x and 4.0.0. 

However, I've also published 5.0.0, which moves away from a tuple return type entirely. In 5.0.0, you will need to directly reference the fields you're interested in.

```python
ext = tldextract.extract(url)
'.'.join((ext.domain, ext.suffix))
```

Issue #305, opened 2023-09-20
Version 3.6.0 introduced a breaking change without bumping to 4.0.0


dror-wing:
Our production code broke after upgrading to version 3.6.0 because of the following line:
```
subdomain, domain, suffix = tldextract.extract(link.lower())
```

We were expecting a breaking change would reflect in a major version bump as per the [semver spec](https://semver.org/).

[Thank you](https://xkcd.com/2347/)

john-kurkowski:
Thanks for letting us know about this issue and your use case. Thank you for the xkcd. 😊

There was some discussion of the possibility of breaking in #300. We wanted to add a field to this project, which sounds like a minor version bump. However, maybe when a public interface is a _tuple_, there's no way to lengthen the tuple (nor shorten it) without it being a breaking change?

In the meantime, for anybody encountering this issue, to work around: directly reference the fields you're interested in.

https://github.com/john-kurkowski/tldextract/blob/f4ea7140e8c893ec8d963293a7e1097177bbc76e/README.md?plain=1#L37-L39

TTycho:
Ouch, this break cases like:
`'.'.join(part for part in ext[-2:] if part)`
I upgraded to direct field referencing but I would consider this a breaking change.

john-kurkowski:
* 3.6.0 has been yanked and re-released as 4.0.0, for people who upgraded their code to rely on 3.6.0's 4-tuple `(subdomain, domain, suffix, is_private)` return type. Lesson learned that increasing the length of a tuple should be a breaking change.
* 5.0.0 has also been released, which migrates away from the tuple entirely, to avoid breaking changes when adding fields in the future. In 5.0.0, you _must_ directly reference the fields you're interested in. No more iterating/indexing/slicing/unpacking the result object returned by this library. For example, `ext[-2:]` will no longer work at all. Use `(ext.domain, ext.suffix)` instead.

PR #306, opened 2023-09-21
Migrate ExtractResult from namedtuple to dataclass

john-kurkowski:
The bugfix in #300 was not the breaking change I wanted. It resulted in #305. To properly allow the original bugfix, add flexibility to the library to e.g. add more metadata fields going forward, and really warrant a major version bump, this PR changes the type of this library's core type `ExtractResult` from `namedtuple` to `dataclass`. I'm considering yanking version 3.6.0 and making _this_ PR's change the 4.0.0.

# Open Issues

- [ ] Does this greatly affect space?
- [ ] Does this greatly affect speed?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,20 +31,6 @@ ExtractResult(subdomain='forums', domain='bbc', suffix='co.uk', is_private=False
 ExtractResult(subdomain='www', domain='worldbank', suffix='org.kg', is_private=False)
 ```
 
-`ExtractResult` is a namedtuple, so it's simple to access the parts you want.
-
-```python
->>> ext = tldextract.extract('http://forums.bbc.co.uk')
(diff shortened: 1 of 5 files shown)


john-kurkowski:
Re: space, at least in the test suite, not necessarily production usage, in main_test.py, memory usage does go up 8%, from 34MB to 37MB on average.

john-kurkowski:
Re: speed, using `timeit` and a hardcoded domain string, I see execution time go _down_ 2-5% on CPython 3.11.

john-kurkowski:
I want this change for future compatibility. I'm not seeing many downsides, besides the breaking change of no longer allowing accessing indexing and slicing tuple members, as there is no longer a tuple. This will be marked a breaking change.

PR #314, opened 2023-11-11
Tests on Windows + GitHub Actions

elliotwutingfeng:
Tests on Windows via GitHub Actions were not working. While trying to fix them, I've uncovered some strange behaviors. Currently all tests pass on macOS, Windows, and Linux for python 3.8-3.12 + pypy3.8.

#### 1. Incomplete path conversion on Windows

```python
import os
from pathlib import Path, PureWindowsPath

os.path.join('/home/anonymous', 'apple')
'/home/anonymous\\apple'

str(Path('/home/anonymous','apple'))
# you can simulate windows style Path() on UNIX systems via str(PureWindowsPath('/home/anonymous','apple'))
'\\home\\anonymous\\apple'
```

Unlike `str(pathlib.Path)`, `os.path.join` does not convert existing forward-slashes, causing incomplete conversion of the monkeypatched UNIX-style paths to Windows-style.

```python
  monkeypatch.setenv("HOME", "/home/john")
  monkeypatch.setenv("XDG_CACHE_HOME", "/my/alt/cache")
  monkeypatch.delenv("TLDEXTRACT_CACHE", raising=False)
```

---

#### 2. Python 3.11 and 3.12 on Windows

For some unknown reason, in Python 3.11 & 3.12 on Windows (but not UNIX), the following receives a [WindowsPath](https://github.com/elliotwutingfeng/tldextract/actions/runs/6809342643/job/18515429911#step:5:163) instead of a string. So I modified `evil_unlink` to perform relative path checks for both string and Path (since `os.unlink` supports both datatypes).

```python
def evil_unlink(filename: str) -> None:
```

---

#### 3. Erroneous stdout on pypy3.9 and pypy3.10 (all OSes)

See <https://github.com/elliotwutingfeng/tldextract/actions/runs/6810665611/job/18519409296#step:5:60>

```python
E       AssertionError: assert '/Users/runne...s bbc co.uk\n' == ' example com...s bbc co.uk\n'
E         + /Users/runner/work/tldextract/tldextract/.tox/pypy39/lib/pypy3.9/site-packages/certifi/cacert.pem None
E            example com
E            bbc co.uk
E           forums bbc co.uk
```

I believe this has to do with the requests library, but I am unable to suppress the output (some kind of tuple?). This is probably the most important issue since pypy only supports the latest 1 or 2 versions of pypy.

---

#### 4. Consider switching to GitHub Actions from Travis

If you are spending money on Travis, you may want to consider switching to GitHub Actions as it is free for public projects like this.

---


tests/test_parallel.py
```diff
@@ -43,9 +44,23 @@ def test_cache_cleared_by_other_process(
     extract("google.com")
     orig_unlink = os.unlink
 
-    def evil_unlink(filename: str) -> None:
+    def is_relative_to(path: Path, other_path: str | Path) -> bool:
+        """Return True if path is relative to other_path or False.
+
+        Taken from standard library (Python 3.9+ required).
```
john-kurkowski, line 50: > Python 3.9+ required

This hunk works even though this project supports Python 3.8+?

john-kurkowski (commented):
Hey! This PR makes sense. Love the path fixes. 

As for CI, I was not committed to Travis CI. I was on their free tier for OSS. However, I have been using GitHub Actions in other projects, and love the integration and community. Happy to switch. 

Before we merge, I would love to see a GitHub Actions checkmark on this PR. Not sure why it's not running.

john-kurkowski:
Okay there's a green checkmark on eb99020. 👍

john-kurkowski (approved):

tests/test_parallel.py
elliotwutingfeng: It's a [standard library method](https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.is_relative_to) introduced in Python 3.9. So I had to implement it here as a function because Python 3.8 does not have it.

john-kurkowski: Gotcha! I clarified the comment a bit.

PR #318, opened 2024-02-22
Remove socket.inet_pton

### LINKED ISSUE #317: platform-dependent parsing of IP addresses
Try the following in MacOS (Note the leading `0` on the last segment of the IP address):
```python
>>> import tldextract
>>> tldextract.extract('127.0.0.01')
ExtractResult(subdomain='', domain='127.0.0.01', suffix='', is_private=False)
>>>
>>> from urllib.parse import urlparse
>>> urlparse('127.0.0.01')
ParseResult(scheme='', netloc='', path='127.0.0.01', params='', query='', fragment='')
```

And then in Linux:
```python
>>> import tldextract
>>> tldextract.extract('127.0.0.01') # note difference in output here
ExtractResult(subdomain='127.0.0', domain='01', suffix='', is_private=False)
>>> 
>>> from urllib.parse import urlparse
>>> urlparse('127.0.0.01') # urllib remains consistent in Linux and Mac
ParseResult(scheme='', netloc='', path='127.0.0.01', params='', query='', fragment='')
```

Several languages recently started to correctly validating these IP addresses (there are security concerns associated to this, e.g., https://github.com/advisories/GHSA-38h6-vxp4-qxvm). 

Environment:
* Python 3.11.4
* tldextract version `5.1.1`
* Linux Debian 5.10.197-1
* MacOS Sonoma

elliotwutingfeng:
Closes  #317

### Changes

- Remove **socket.inet_pton** which does not reject IPv4 addresses and IPv6 (dual) addresses with leading IPv4 zeroes  on macOS.
- Add **test_looks_like_ipv6**

### Unresolved

I've also run [checks](https://github.com/elliotwutingfeng/tldextract/actions/runs/8000214791/job/21849274115) on Python 3.8.7 which was affected by the leading zeroes inconsistency in the standard library method **ipaddress.IPv4Address** (see #292). It turns out **ipaddress.IPv6Address** is also affected by the same inconsistency (for trailing IPv4 portion).

I would propose copying the implementation of **ipaddress.IPv6Address** or some other implementation like the one in [Go](https://github.com/golang/go/blob/master/src/net/netip/netip.go), modifying it to avoid the leading zeroes bug even on older Python versions, and removing portions that we don't need, like checks if input is of `str` type, though it means we will have to ensure that our implementation is up-to-date with upstream. Upside is we will have more control over parsing bugs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
         os: [macos-latest, windows-latest, ubuntu-latest]
         language:
           [
-            {python-version: "3.8", toxenv: "py38"},
+            {python-version: "3.8.1", toxenv: "py38"},
             {python-version: "3.9", toxenv: "py39"},
             {python-version: "3.10", toxenv: "py310"},
             {python-version: "3.11", toxenv: "py311"},
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,10 +5,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-20.04]
+        os: [macos-latest, windows-latest, ubuntu-latest]
         language:
           [
-            {python-version: "3.8.7", toxenv: "py38"},
+            {python-version: "3.8", toxenv: "py38"},
             {python-version: "3.9", toxenv: "py39"},
             {python-version: "3.10", toxenv: "py310"},
             {python-version: "3.11", toxenv: "py311"},
@@ -16,11 +16,11 @@ jobs:
             {python-version: "pypy3.8", toxenv: "pypy38"},
           ]
         include:
-          - os: ubuntu-20.04
+          - os: ubuntu-latest
             language: {python-version: "3.8", toxenv: "codestyle"}
-          - os: ubuntu-20.04
+          - os: ubuntu-latest
             language: {python-version: "3.8", toxenv: "lint"}
-          - os: ubuntu-20.04
+          - os: ubuntu-latest
             language: {python-version: "3.8", toxenv: "typecheck"}
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest]
+        os: [macos-latest, windows-latest, ubuntu-20.04]
         language:
           [
             {python-version: "3.8.7", toxenv: "py38"},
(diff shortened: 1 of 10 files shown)


.github/workflows/ci.yml
```diff
@@ -5,10 +5,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-latest, windows-latest, ubuntu-latest]
+        os: [macos-latest, windows-latest, ubuntu-20.04]
```
john-kurkowski, line 8: What does pinning only Ubuntu do?

.github/workflows/ci.yml
```diff
@@ -5,10 +5,10 @@ jobs:
...
         language:
           [
-            {python-version: "3.8", toxenv: "py38"},
+            {python-version: "3.8.7", toxenv: "py38"},
```
john-kurkowski, line 11: Does the new test suite pass for patch versions newer than 3.8.7, like 3.8.8?

tldextract/remote.py
```diff
@@ -3,16 +3,9 @@
 from __future__ import annotations
 
 import re
-from collections.abc import Callable
 from ipaddress import AddressValueError, IPv6Address
 from urllib.parse import scheme_chars
 
-inet_pton: Callable[[int, str], bytes] | None
-try:
-    from socket import AF_INET, AF_INET6, inet_pton  # Availability: Unix, Windows.
```
john-kurkowski, line 12: I forget, was `inet_pton` for a faster implementation? This PR sacrifices those for correctness?

john-kurkowski (commented):
I'm loving the shorter code! If we can fix/skip CI, I'm tempted to accept the PR as-is, and call it a day. I'm wary of taking on IP parsing code, when a fix is already available in years-old Python patch releases.

.github/workflows/ci.yml
elliotwutingfeng: To allow me to run CI on an old version of Python 3.8. It is temporary.

.github/workflows/ci.yml
elliotwutingfeng: According to Python docs under ipaddress module, for Python 3.8, the ipaddress bug was fixed starting from 3.8.12.

tldextract/remote.py
elliotwutingfeng: Yes, it was for faster implementation, but its behavior is platform specific. We just learned that macOS still allows leading zeroes in ipv4 addresses today.

elliotwutingfeng:
GitHub Action **setup-python** for Windows does not have Python 3.8.12 or later.

john-kurkowski:
Yeah, skip it. Maybe it can be a single version `>= 3.8.12` check? That is, platform-independent?

tldextract/cache.py
```diff
@@ -38,8 +38,7 @@ def md5(*args: bytes) -> hashlib._Hash:
 
 
 def get_pkg_unique_identifier() -> str:
-    """
-    Generate an identifier unique to the python version, tldextract version, and python instance.
+    """Generate an identifier unique to the python version, tldextract version, and python instance.
```
john-kurkowski, line 41: Do you know what generated these linebreak diffs? I'm just curious, to reduce noise in our tooling. The latest version of `black .` and `ruff format .` don't seem to care about this change either way.

john-kurkowski:
Thanks for hopping on this!

elliotwutingfeng:
Python versions where the **ipaddress** standard library erroneously accepts leading IPv4 zeroes in IPv4 addresses / IPv6 (dual) addresses.

- Python **3.8.0** to **3.8.11** inclusive
- Python **3.9.0** to **3.9.4** inclusive

Reference: <https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Address> (referenced hyperlink is for **ipaddress.IPv4Address** but applies to ipaddress.IPv6Address dual addresses too)

PR #324, opened 2024-03-11
Switch from pip to uv, during tox

john-kurkowski:
Significantly speeds up fresh tox environments, like in CI. Does not require contributors to use uv directly, outside of tox.

# Open Issues

* [ ] How much do we like a new tool that's at version ~~0.1.16~~ 0.2.3? 😅
* [x] Await fix or work around PyPy tests failing on Windows: astral-sh/uv#2316

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
           check-latest: true
       - name: Install Python requirements
         run: |
-          pip install --upgrade --system tox tox-uv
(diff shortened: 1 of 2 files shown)


john-kurkowski:
> How much do we like a new tool that's at version ~~0.1.16~~ 0.2.3?

I think the main risk is failing some CI runs. If the tool becomes a thorn in CI's side, this commit is easily reverted. The risk to contributors is low. The risk to users is very low.

PR #327, opened 2024-03-19
Keep more of GitHub release notes body

john-kurkowski:
Only omit the first paragraph, "What's Changed." Keep the e.g. "New Contributors" paragraph and "Full Changelog" link.

### DIFF
--- a/scripts/release.py
+++ b/scripts/release.py
@@ -81,7 +81,11 @@ def verify_build(is_test: str) -> None:
 
 
 def generate_github_release_notes_body(token: str, version: str) -> str:
-    """Generate and grab release notes URL from Github."""
+    """Generate and grab release notes URL from Github.
+
+    Delete their first paragraph, because we track its contents in a tighter
+    form in CHANGELOG.md. See `get_changelog_release_notes`.
+    """
     response = requests.post(
         "https://api.github.com/repos/john-kurkowski/tldextract/releases/generate-notes",
         headers={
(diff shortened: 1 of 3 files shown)


john-kurkowski:
FYI @ekcorso

scripts/release.py
```diff
@@ -131,25 +124,15 @@ def get_changelog_release_notes(release_notes_url: str, version: str) -> str:
     if match:
         return str(match.group(1)).strip()
     else:
-        print(
-            f"WARNING: Failed to parse changelog release notes. Manually copy this version's notes from the CHANGELOG.md file to {release_notes_url}.",
```
john-kurkowski, line 135: Haven't tripped on this error handler yet, but I think the URL in it was wrong. It's a _compare_ link like https://github.com/john-kurkowski/tldextract/compare/5.1.1...5.1.2, not a create-release link. I moved this error print to after we have the release link, with a hardcoded URL.

scripts/release.py
```diff
@@ -100,24 +104,13 @@ def generate_github_release_notes_body(token: str, version: str) -> str:
...
-        return match.group(1)
-    else:
-        print(
-            "WARNING: Failed to parse release notes URL from GitHub response.",
-            file=sys.stderr,
-        )
-        return ""
+    body = str(response.json()["body"])
+    paragraphs = body.split("\n\n")
+    return "\n\n".join(paragraphs[1:])
```
ekcorso, line 110: So concise and it got rid of the REGEX-- love it. Still a little brittle if this format changes, but I think this is a great improvement.

scripts/release.py
ekcorso: Yeah, good point. I think I garbled this warning when I added the `release_notes_url`. Your new placement is correct: it should be directing the user to the url of the newly created draft release to manually add the notes. Thanks for catching this!

ekcorso (approved):

Issue #342, opened 2025-03-05
Reverse Domain Name Notation?


jvanasco:
I used tldextract to implement a (slight) variant of Reverse Domain Name notation, which is often used to handle namespace issues -- or simply better sorting of large lists of domains.  (see https://en.wikipedia.org/wiki/Reverse_domain_name_notation )

Instead of relying on strict reverse dns, I use TLD extract to pull out the "registered" domain and suffix,  so sorting happens on `.co.uk` instead of `.uk.co`.

The code is below, and I'd be happy to submit a PR if you think it's useful to more people.

```python
def reverse_domain_name(domain):
    result = tldextract.extract(domain)
    stack =[result.suffix, result.domain]
    if result.subdomain:
        stack.extend(reversed(result.subdomain.split(".")))
    return ".".join(stack)

print(reverse_domain_name("example.co.uk"))
# co.uk.example

print(reverse_domain_name("login.example.co.uk"))
# co.uk.example.login

print(reverse_domain_name("example.com"))
# com.example

print(reverse_domain_name("dash.example.com"))
# com.example.dash

print(reverse_domain_name("login.dash.example.com"))
# com.example.dash.login
```

john-kurkowski:
Nice, yeah, I'd be open to that new property.

PR #345, opened 2025-03-25
Add reverse domain name notation property

jvanasco:
A reverse domain notation function, as offered in #342

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ After upgrading, update your cache file by deleting it or via `tldextract
 ## unreleased
 
 * Misc.
-  * Added function `reverse_domain_name` to implement Reverse Domain Name Notation.
+  * Add `reverse_domain_name` property, for Reverse Domain Name Notation.
     See:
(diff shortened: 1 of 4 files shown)


john-kurkowski:
Thank you! I was picturing this as a new property. Rather than expand the entrypoints into this library. See cff0327. What do you think?

john-kurkowski:
Thanks for the contribution!

jvanasco:
You're welcome!

FYI, I realized last night that I greatly prefer your idea to implement this as a property of ExtractResult.  This isn't a true reverse domain notation - it's a slight adaptation of the concept (although it is what most people implement IRL) based on the TLD.  As such, it makes much more sense as a property of the ExtractResult than a standalone function.

john-kurkowski:
> This isn't a true reverse domain notation - it's a slight adaptation of the concept (although it is what most people implement IRL) based on the TLD.

Good catch. I noted this in ffd6203d7bc51e72bc3891f352e9b7dd71b7ec06.

PR #347, opened 2025-04-08
Deprecate `registered_domain` result property

### LINKED ISSUE #138: Semantics of `registered_domain` property for private domains

john-kurkowski:
Use `top_domain_under_public_suffix` instead, which has the same behavior but a more accurate name.

Closes #138.

### DIFF
--- a/tldextract/tldextract.py
+++ b/tldextract/tldextract.py
@@ -38,6 +38,7 @@ from __future__ import annotations
 
 import os
 import urllib.parse
+import warnings
 from collections.abc import Collection, Sequence
 from dataclasses import dataclass, field
 from functools import wraps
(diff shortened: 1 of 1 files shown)


jvanasco:
For pinning purposes, `registered_domain` will be around until <7.0.0 -- correct?

jvanasco:
I just saw the changelog...

I think there is an error in the docstring:

    -        .. deprecated:: 6.0.0
    +        .. deprecated:: 5.3.1
    +        .. versionremoved:: 6.0.0
    

versionremoved

john-kurkowski:
Good catch. Copypasta. Fixed in 4285fc0a9b54f6ebcf0e4e19c93cd1f12c1336d1.

PR #348, opened 2025-08-01
Restrict CI workflow access privileges to read-only

elliotwutingfeng:
GitHub security recommends setting permissions for this workflow to read-only since write access is not being used.


john-kurkowski:
Least privilege principle. Thank you!

PR #355, opened 2025-10-20
Block network requests during tests

john-kurkowski:
For the test suite to explicitly document the behavior of this library: when does the library reach out to remote servers?

# Draft

- [x] Merge latest upstream
- [ ] Explicitly stub network in test cases that don't require network
- [ ] Pass test cases that require network; pass them through the network block (do any test cases truly require it?)


P4X-ng (approved):

john-kurkowski:
This is stale and not urgent. If anybody has time to pick it up, I'd be interested.

PR #362, opened 2026-04-01
Fix README: extract_urllib is on the TLDExtract instance, not the module

danishashko:
The README example under "How to validate URLs before extraction" calls:

```python
result = tldextract.extract_urllib(split_url)
```

But `extract_urllib` doesn't exist at module level. It's a method on `TLDExtract` instances. The module already exports `extract = TLDExtract()` as a ready-to-use instance, so the correct call is:

```python
result = tldextract.extract.extract_urllib(split_url)
```

Running the original example raises `AttributeError: module 'tldextract' has no attribute 'extract_urllib'`.


elliotwutingfeng:
Hi, thanks for the catch, however, on 5.3.1, `result = tldextract.extract.extract_urllib(split_url)` did not work for me, but this does

```python
result = tldextract.TLDExtract().extract_urllib(split_url)
# ExtractResult(subdomain='example', domain='com:8080', suffix='', is_private=False)
```

john-kurkowski:
Thanks y'all!

PR #368, opened 2026-06-17
Fix reverse_domain_name emitting stray dots for empty components

### LINKED ISSUE #370: reverse_domain_name emits a spurious leading dot for suffixless inputs (IPs, unknown TLDs)
### Summary
`ExtractResult.reverse_domain_name` joins `[suffix, domain, subdomain...]` without dropping empty components, so when `suffix == ""` (IP addresses, unknown/invalid TLDs) the result carries a spurious leading dot. The sibling `fqdn` property filters out empty components; `reverse_domain_name` does not.

### Reproduction
```python
import tldextract
print(tldextract.extract("http://127.0.0.1/").reverse_domain_name)               # '.127.0.0.1'
print(tldextract.extract("http://google.notavalidsuffix/").reverse_domain_name)  # '.notavalidsuffix.google'
```

### Expected
No leading dot: `'127.0.0.1'` and `'notavalidsuffix.google'`.

### Actual
A leading `.` is prepended on every suffixless input.

### Fix sketch
Filter empty components before joining, matching the `fqdn` property's behavior.

### Environment
tldextract 5.3.2.dev (master @ f06eec7), Python 3.12.

gaoflow:
Fixes #370.

### Problem

`ExtractResult.reverse_domain_name` builds its label stack with `[self.suffix, self.domain]` unconditionally, so when either is empty the joined output gets a stray dot. This happens for very common inputs:

```python
>>> import tldextract
>>> tldextract.extract("localhost").reverse_domain_name
'.localhost'
>>> tldextract.extract("127.0.0.1").reverse_domain_name
'.127.0.0.1'
>>> tldextract.extract("com").reverse_domain_name
'com.'
>>> tldextract.extract("").reverse_domain_name
'.'
```

Reverse-domain-name notation is a sequence of dot-separated labels; a leading dot, trailing dot, or `..` is never valid. The sibling `fqdn` and `top_domain_under_public_suffix` properties already filter empty components before joining, so this property is just inconsistent with them.

### Fix

Filter out empty components, using the same `[i for i in (...) if i]` pattern already used by `fqdn`:

```python
stack = [i for i in (self.suffix, self.domain) if i]
```

Now `localhost` → `'localhost'`, `127.0.0.1` → `'127.0.0.1'`, `com` → `'com'`, `''` → `''`, and ordinary domains are unchanged (`login.example.co.uk` → `'co.uk.example.login'`).

### Tests

Added `test_reverse_domain_name_notation_no_suffix` (and a doctest) covering empty suffix, IP, suffix-only, and empty input. It fails before the change (`'.localhost' != 'localhost'`) and passes after. Full suite: 82 passed; `ruff check`, `ruff format --check`, and `mypy` all clean.

I used an AI assistant to help investigate this under my direction, and I have reviewed and verified the change myself.


john-kurkowski:
Thank you!

PR #369, opened 2026-06-19
Normalize punycode labels in extraction results

### LINKED ISSUE #366: Punycode/Unicode form returns different suffix/fqdn for the same DNS domain — enables blocklist/allowlist bypass
### Description

`tldextract.extract()` returns **different output** for the same DNS domain depending on whether the input uses Unicode (IDN) or punycode (ACE) encoding. Applications that use `tldextract` for security-relevant domain checks (allowlists / blocklists / SSO origin validation / phishing detection) can be bypassed by submitting the alternate encoding form.

### Reproducer (tldextract 5.3.0, also reproduces in current main)

```python
import tldextract

extractor = tldextract.TLDExtract(cache_dir=None, suffix_list_urls=(), fallback_to_snapshot=True)

# Same DNS domain, two valid representations:
r_uni  = extractor("test.公司.cn")
r_puny = extractor("test.xn--55qx5d.cn")    # xn--55qx5d == 公司 (ACE)

print(r_uni.suffix,  "|", r_uni.fqdn)        # 公司.cn  | test.公司.cn
print(r_puny.suffix, "|", r_puny.fqdn)       # xn--55qx5d.cn | test.xn--55qx5d.cn

assert r_uni.suffix != r_puny.suffix    # security check fails
assert r_uni.fqdn   != r_puny.fqdn
```

Other affected suffixes: any IDN ccTLD — `рф / xn--p1ai`, `ישראל / xn--4dbrk0ce`, `中國 / xn--fiqz9s`, `сайт.рф / xn--80aswg.xn--p1ai`, etc.

### Threat model

```python
BLOCKED_SUFFIXES = {"公司.cn"}                       # admin configures in Unicode
def is_blocked(url): return extractor(url).suffix in BLOCKED_SUFFIXES

is_blocked("https://attacker.公司.cn/")             # True  — correctly blocked
is_blocked("https://attacker.xn--55qx5d.cn/")       # False — BYPASS
```

The reverse direction is also broken: an allowlist stored in punycode is bypassed by Unicode input.

Use cases where this matters in the wild include phishing filters, OAuth redirect-URI validation, SSO origin checking, ingress firewalls / WAFs, ad blockers, and email scanners. Many of those configure their lists in human-readable Unicode but receive URLs in either form; DNS treats the two forms as identical.

### Suggested fix

Either:
- **(A, recommended)** Canonicalize the input host to a single form (punycode/ACE) before matching against the Public Suffix List, and return both canonical and original on the result (`suffix` and `suffix_unicode`, or similar). This makes the default safe.
- **(B)** Document loudly that callers must normalize themselves before using `suffix`/`fqdn`/`registered_domain` for security comparisons, and ship helper `.canonical()` / `.normalized()` properties.

binggao1230:
Summary:
- Decode valid `xn--` labels before assembling extraction result fields so Unicode and ACE forms produce the same suffix/fqdn.
- Preserve invalid punycode labels unchanged.
- Add regression coverage for equivalent Unicode and punycode IDN suffixes.

Tests:
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 .tox/py310/bin/python -m pytest tests/main_test.py -k puny -q`
- `.tox/py310/bin/ruff format .`
- `uv run tox -e py310 -- -p no:gitignore`
- `uv run tox -e lint`
- `uv run tox -e typecheck`
- `uv run tox -e codestyle`

Note: the local fresh tox environment installs pytest 9.1.1 and pytest-gitignore currently fails during plugin loading, so the py310 run disables only that plugin with `-p no:gitignore`.

Fixes #366

AI assistance was used under my direction.


john-kurkowski:
Thanks for the contribution. I’m not going to merge this approach. It changes the representation of every valid ACE label in returned results, breaking established behavior.

Decoding ACE labels also does not provide the canonicalization guarantee described in #366. Raw Unicode input remains unvalidated and unnormalized, while other equivalent inputs such as differently cased ASCII names can still produce different strings. A security-sensitive caller would still need a separate hostname normalization step.

I’ve resolved #366 via documentation. That preserves the current API contract and raises awareness of the string comparison issue.

PR #372, opened 2026-07-26
Add opt-in PSL default wildcard rule

### LINKED ISSUE #338: Default wildcard rule missing from the PSL algorithm implementation
The [PSL formal algorithm](https://github.com/publicsuffix/list/wiki/Format#algorithm) includes the following step:

> - If no rules match, the prevailing rule is "*".

This rule means that if an explicit public suffix can't be found for a domain in the PSL, the (actual, not effective) TLD is treated as a public suffix. This allows new TLDs (or custom internal TLDs) to be handled without requiring an explicit update to the PSL.

(It also means that the only time the PSL algorithm doesn't return a registrable domain is when the domain is itself a public suffix, which helps disambiguate that case.)

The PSL project has tests for this in the [standard test suite](https://github.com/publicsuffix/list/blob/master/tests/test_psl.txt):
```
// Unlisted TLD.
checkPublicSuffix('example', null);
checkPublicSuffix('example.example', 'example.example');
checkPublicSuffix('b.example.example', 'example.example');
checkPublicSuffix('a.b.example.example', 'example.example');
```

Here's the output for the test domains in `tldextract` 5.1.2:
```
>>> tldextract.extract('example').registered_domain                                                                                                                                                                                                                          
''
>>> tldextract.extract('example.example').registered_domain
''
>>> tldextract.extract('b.example.example').registered_domain
''
>>> tldextract.extract('a.b.example.example').registered_domain
''
```

The last three results are turning `''`, when they should return `example.example`.

deepakganesh78:
Fixes #338.

The PSL algorithm specifies `*` as the prevailing rule when no explicit rule matches, but enabling that behavior by default would change how callers distinguish recognized suffixes from internal or invalid hostnames.

This adds an opt-in `include_psl_default_rule` setting while preserving the current default behavior. The setting is available on `TLDExtract`, individual extraction calls, the top-level `extract()` helper, and the CLI. When enabled, unlisted TLDs use their final label as the public and registry suffix, matching the PSL standard test cases.

Validation:
- `tox -e py310 -- -q` — 87 passed, 1 skipped
- `tox -e lint` — passed
- `tox -e typecheck` — passed
- targeted main and CLI tests — 50 passed
- Ruff format/check passed for all touched Python files

The test environment constrained pytest to `<9` because the repository's existing `pytest-gitignore` dependency is not compatible with pytest 9.


john-kurkowski:
Thanks for the contribution and preserving existing behavior. I decided in #338 to not support the PSL default fallback and to document a caller-side recipe, instead of adding new API and runtime modes. Declining this PR to match.

PR #378, opened 2026-09-16
docstring changes for api changes

jvanasco:
I needed to update a package that relies on TLDExtract and tried to figure out how to best handle the API changes with `registered_domain`.

While doing this, I realized 2 things:
* the docstrings don't note the version added for `top_domain_under_public_suffix` and `top_domain_under_registry_suffix`
* the docstrings deprecate `registered_domain` in 5.3.2, but it was deprecated in the 5.3 release

I'm not sure if ":added" or ":versionadded" is correct.

Also a slight note: there hasn't been a 'tag' since version 2.0.2;  if you were still settings tags, but haven't pushed tags to GitHub, it would be awesome if you did a push.  If you stopped tagging, it would be great if you started again.  It's so much easier to figure out dependency issues when there are tags.


john-kurkowski:
Thanks for catching these. `.. versionadded::` is the right directive.

As for past tags, they're all there already. Check out https://github.com/john-kurkowski/tldextract/tags . If you're not seeing them, a local clone may need `git fetch --tags`.

jvanasco:
>  as for past tags, they're all there already. Check out https://github.com/john-kurkowski/tldextract/tags 

This is odd. There must have been an issue with GitHub when I filed this, because it was only showing tags through the 2.0 release (or around there).  I didn't even look at my git installation, just the GitHub interface.

